PR #44, opened 2023-05-16
refactor: Update xapi_events_all_parsed and supporting view

bmtcril:
- Extract org and object id into the top level of xapi_events_all_parsed.
- Use the String type instead of JSON for the event JSON data, since the JSON type is so poorly supported.
- Add time_last_dumped to event_sink.relationships PK to help speed up queries on that data

### DIFF
--- a/tutoroars/templates/oars/jobs/init/oars_init_schemas_tables_users.sh
+++ b/tutoroars/templates/oars/jobs/init/oars_init_schemas_tables_users.sh
@@ -4,15 +4,24 @@ clickhouse client --user "{{ CLICKHOUSE_ADMIN_USER }}" --password="{{ CLICKHOUSE
 SET allow_experimental_object_type=1;
 
 -- Create various non-admin users reporting users
-CREATE USER IF NOT EXISTS {{ OARS_CLICKHOUSE_LRS_USER}} IDENTIFIED WITH sha256_password BY '{{ OARS_CLICKHOUSE_LRS_PASSWORD }}';
-CREATE USER IF NOT EXISTS {{ OARS_CLICKHOUSE_REPORT_USER}} IDENTIFIED WITH sha256_password BY '{{ OARS_CLICKHOUSE_REPORT_PASSWORD }}';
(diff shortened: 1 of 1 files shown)


bmtcril:
@Ian2012 @mariajgrimaldi @soryrawyer I can't ping you all for reviews directly, but would appreciate your eyes on this. It should tidy up some longstanding issues with this table, but once merged will require you to drop the `xapi_events_all_parsed` table and `xapi_events_all_parsed_mv` view so they can be recreated. Hopefully that will be the last time we need to do this, but I kinda doubt it.

tutoroars/templates/oars/jobs/init/oars_init_schemas_tables_users.sh
```diff
@@ -31,11 +40,12 @@ PRIMARY KEY (emission_time, event_id);
...
     event_id UUID NOT NULL,
     verb_id String NOT NULL,
-    actor_id UUID NOT NULL,
+    actor_id String NOT NULL,
+    object_id String NOT NULL,
     org String NOT NULL,
     course_id String NOT NULL,
     emission_time DateTime64(6) NOT NULL,
-    event JSON NOT NULL
+    event_str String NOT NULL
```
pomegranited, line 48: I don't understand how  the JSON data type is poorly supported in Clickhouse? I've been loving how I can dig into it with the SQLExplorer in Superset, and it makes it very easy to dig out event specifics that aren't in their own fields (like whether a problem was answered successfully or not).

pomegranited: Ok, sure.. that's still a shame! I had hoped we could avoid many tedious schema changes whenever we want to chart new xAPI event types! But oh well..

I guess anyone who's in a rush can still use the https://clickhouse.com/docs/en/sql-reference/functions/json-functions

pomegranited (approved):
:+1: LGTM

- [x] I tested this by spinning up a fresh OARS stack, running ` tutor local do load-xapi-test-data`, and checking the xapi tables are populated as expected. Also ensured that new events are still percolating through from the LMS.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] <strike>Includes documentation</strike> N/A

PR #54, opened 2023-06-07
feat: Add Vector for tracking logs and xapi

bmtcril:
This pulls in Cairn's Vector implementation and adds an xAPI layer on top of it. It currently does not have all of the Cairn migrations as we're looking at moving to an Alembic solution, but should allow us to run xAPI and all downstream dbt / Superset without an LRS by just changing the database (since the top level source table name and structure are "the same enough"). They can also be run side-by-side and both capture the same events.

The xAPI logging can be turned off on the event-routing-backends / LMS configuration, effectively disabling that functionality for performance. Right now I don't think there's a way to turn off tracking logs without turning off RUN_VECTOR, and I'm not sure if that's desirable. 🤔 

There are some naming inconsistencies to keep both the Cairn and OARS naming structures for their tables, which are probably not necessary long term but I wanted to keep this PR relatively contained.


tutoroars/plugin.py
```diff
@@ -33,6 +33,8 @@
         ("OARS_EVENT_SINK_RELATIONSHIPS_TABLE", "course_relationships"),
         ("OARS_EVENT_SINK_OVERVIEWS_TABLE", "course_overviews"),
         ("OARS_EVENT_SINK_CLICKHOUSE_TIMEOUT_SECS", "5"),
+        # ClickHouse tracking log settings
+        ("OARS_TRACKING_LOG_DATABASE", "openedx"),
```
bmtcril, line 37: This may not be the name we want to settle on given the pattern of the xapi database being separate, but I wanted to change as little as possible in initial testing.

tutoroars/plugin.py
```diff
@@ -43,6 +45,9 @@
                 "edx-event-routing-backends==5.3.1",
             ],
         ),
+        # Vector settings
+        ("VECTOR_DOCKER_IMAGE", "docker.io/timberio/vector:0.30.0-alpine"),
```
bmtcril, line 49: This is a version bump from what Cairn uses so we could get the `date_time_best_effort` setting in the ClickHouse sink to simplify the parsing of more exact dates.

tutoroars/templates/oars/apps/vector/local.toml
```diff
@@ -0,0 +1,16 @@
...
+
+### Sources
+# Capture logs from all docker containers
+[sources.docker_logs]
+type = "docker_logs"
+[transforms.openedx_containers]
+type = "filter"
+inputs = ["docker_logs"]
+condition = 'includes(["lms", "cms", "lms-worker"], .label."com.docker.compose.service")'
+[transforms.lms_worker]
```
bmtcril, line 11: It's likely this isn't needed and we can just use `openedx_containers`, but during testing the xapi regex failures were so numerous it made developement impossible.

tutoroars/templates/oars/apps/vector/partials/common-post.toml
```diff
@@ -0,0 +1,114 @@
...
+if err_timestamp != null {
+  log("Unable to parse timestamp from tracking log 'time' field: " + err_timestamp, level: "warn")
+  time, err_timestamp = parse_timestamp(parsed_json.timestamp, "%+")
+  if err_timestamp != null {
+    log("Unable to parse timestamp from tracking log 'timestamp' field: " + err_timestamp, level: "error")
+    abort
+  }
+}
+event_id = parsed_json.id
+. = {"event_id": event_id, "emission_time": format_timestamp!(time, format: "%+"), "event_str": encode_json(parsed_json)}
```
bmtcril, line 70: At the moment this table structure differs from the Ralph table I actually want to write to. I was having trouble writing to the JSON column, so omitted it for this initial testing. This needs to be fixed before merging.

tutoroars/patches/k8s-deployments
```diff
@@ -0,0 +1,115 @@
+---
+####### OARS plugin
+# log collection
+# https://vector.dev/docs/setup/installation/platforms/kubernetes/
+# https://github.com/timberio/vector/blob/master/distribution/kubernetes/vector-agent/resources.yaml
+apiVersion: v1
```
Ian2012, line 6:
```suggestion
---
apiVersion: v1
```

tutoroars/templates/oars/apps/vector/partials/common-post.toml
bmtcril: Or maybe it doesn't, if we don't use the json column for anything else. Ralph uses it, but maybe we don't care?

tutoroars/templates/oars/apps/vector/local.toml
bmtcril: Remove this and am just using `openedx_containers` now.

tutoroars/templates/oars/apps/vector/partials/common-post.toml
Ian2012: We can make the JSON column with a default value of "{}" and based all the work on the other available field

tutoroars/patches/k8s-deployments
```diff
@@ -324,3 +324,121 @@ spec:
...
+            - labelSelector:
+                matchExpressions:
+                - key: app.kubernetes.io/name
+                  operator: In
+                  values:
+                    - lms
+              topologyKey: kubernetes.io/hostname
+      containers:
+        - name: vector
+          image: {{ VECTOR_DOCKER_IMAGE }}
```
Ian2012, line 402:
```suggestion
          image: {{ DOCKER_IMAGE_VECTOR }}
```

Ian2012 (approved):
LGTM, just address the variable naming for the vector image

PR #70, opened 2023-06-20
feat: Add preliminary dbt command support

bmtcril:
Allows us to run the dbt commands using OARS configuration defaults. This currently does a whole build process instead of having it's own image, and uses the head of `main` on https://github.com/openedx/oars-dbt (by default, this is configurable to test branches etc). Neither of these are good long term, but this should help people get started.

`dbt run` will be executed as part of Tutor init.

Test with:
`tutor local do dbt --help `

To see if the project compiles (safe, changes nothing):
`tutor local do dbt -c "compile"`

To run it (generate views, etc) against your Tutor configured ClickHouse:
`tutor local do dbt -c "run"`

Options can also be added, this will run the given model and everything it relies on in dbt, using 4 threads:
`tutor local do dbt -c "run -m +enrollments_by_day --threads 4"`

### DIFF
--- a/tutoroars/templates/oars/apps/scripts/oars/dbt.sh
+++ b/tutoroars/templates/oars/apps/scripts/oars/dbt.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+python3 -m venv virtualenv
+. virtualenv/bin/activate
+
+echo "Installing dbt packages..."
+pip install pip install {{ DBT_PACKAGES }}
+
+echo "Installing oars-dbt"
+git clone {{ DBT_REPOSITORY }}
+
+cd oars-dbt/oars || exit
+
+echo "Installing dbt dependencies"
+dbt deps --profiles-dir /app/oars/scripts/oars/
+
+echo "Running dbt $*"
+dbt "$@" --profiles-dir /app/oars/scripts/oars/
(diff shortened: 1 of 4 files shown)


bmtcril:
@Ian2012 @SoryRawyer  dbt now runs on init, and other commands on demand so this should be ready for review.

Ian2012 (approved):
LGTM

tutoroars/templates/oars/apps/scripts/oars/dbt.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+python3 -m venv virtualenv
+. virtualenv/bin/activate
+
+echo "Installing dbt packages..."
+pip install pip install {{ DBT_PACKAGES }}
```
SoryRawyer, line 7: Is this (and the equivalent line in the init script) supposed to be just `pip install {{ DBT_PACKAGES }}`?

SoryRawyer (commented):
I don't know much about adding tutor commands but the dbt-related potions look good to me!

pomegranited (approved):
:+1: Oo thank you for this @bmtcril ! Now I get to play with dbt!

It'll be much better with a pre-built image, but this is a good start :)

- [x] I tested this using the PR test instructions.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike>
- [x] Includes documentation

PR #78, opened 2023-06-21
feat: CI run tutor local with OARS on PR

bmtcril:


.github/workflows/test_tutor_oars.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          python-version: 3.9
+      - name: Install python reqs
+        run: pip install -r requirements/dev.txt
+      - name: Install pars
+        run: pip install -e .
+      - name: Tutor local launch
+        run: tutor config save
+      - name: Tutor build openedx
+        run: tutor images build openedx
```
Ian2012, line 30: This can be painful, so let's make it require CC approval and configure a local system cache for it.

Also, remember that for dev is `tutor images build openedx-dev`

bmtcril: Yeah this is pretty experimental, just trying to get it to a working state. I hear there are new tutor capabilities that may help this as well, so I'm looking into that next.

Ian2012: Yep, also here are some CI processes we can use:

https://github.com/overhangio/test-docker-build/blob/main/.github/workflows/build.yml

PR #82, opened 2023-06-22
feat!: prefer verb_id for primary key in event MVs (FC-0024)

SoryRawyer:
Removes `event_type` from top-level event MVs and changes the data type of `verb_id` to `LowCardinality(String)`. As a result of the removal, the MV tables now use `verb_id` in the primary key instead of `event_type`. The data type change should be transparent to users of these tables while improving performance.

I've tested this by running `tutor local do init --limit oars` using my local copy of the repo and confirmed data can still be inserted via `tutor local do load-xapi-test-data`.

### DIFF
--- a/tutoroars/templates/oars/jobs/init/clickhouse/oars_init_schemas_tables_users.sh
+++ b/tutoroars/templates/oars/jobs/init/clickhouse/oars_init_schemas_tables_users.sh
@@ -87,8 +87,8 @@ CREATE TABLE IF NOT EXISTS {{ OARS_XAPI_DATABASE }}.{{ OARS_ENROLLMENT_EVENTS_TA
     `verb_id` LowCardinality(String) NOT NULL,
     `enrollment_mode` LowCardinality(String)
 ) ENGINE = MergeTree
-PRIMARY KEY (org, course_id, enrollment_mode)
-ORDER BY (org, course_id, enrollment_mode, emission_time);
+PRIMARY KEY (org, course_id)
+ORDER BY (org, course_id, actor_id, enrollment_mode, emission_time);
 
 -- Materialized view that moves data from the processed xAPI table to
 -- the enrollment events table
--- a/tutoroars/templates/oars/jobs/init/clickhouse/oars_init_schemas_tables_users.sh
+++ b/tutoroars/templates/oars/jobs/init/clickhouse/oars_init_schemas_tables_users.sh
@@ -84,9 +84,8 @@ CREATE TABLE IF NOT EXISTS {{ OARS_XAPI_DATABASE }}.{{ OARS_ENROLLMENT_EVENTS_TA
     `object_id` String NOT NULL,
     `course_id` String NOT NULL,
     `org` String NOT NULL,
-    `verb_id` String NOT NULL,
-    `event_type` String NOT NULL,
-    `enrollment_mode` String
+    `verb_id` LowCardinality(String) NOT NULL,
+    `enrollment_mode` LowCardinality(String)
 ) ENGINE = MergeTree
 PRIMARY KEY (org, course_id, enrollment_mode)
 ORDER BY (org, course_id, enrollment_mode, emission_time);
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @SoryRawyer!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

tutoroars/templates/oars/jobs/init/clickhouse/oars_init_schemas_tables_users.sh
```diff
@@ -84,9 +84,8 @@ CREATE TABLE IF NOT EXISTS {{ OARS_XAPI_DATABASE }}.{{ OARS_ENROLLMENT_EVENTS_TA
...
     `course_id` String NOT NULL,
     `org` String NOT NULL,
-    `verb_id` String NOT NULL,
-    `event_type` String NOT NULL,
-    `enrollment_mode` String
+    `verb_id` LowCardinality(String) NOT NULL,
+    `enrollment_mode` LowCardinality(String)
 ) ENGINE = MergeTree
 PRIMARY KEY (org, course_id, enrollment_mode)
 ORDER BY (org, course_id, enrollment_mode, emission_time);
```
bmtcril, line 91: Unrelated to this PR, but is there a reason why we're not using the actor id in the key / order?

SoryRawyer: My reason for leaving it out of the key is that the recommendation I've seen from the ClickHouse docs is to prefer sparser primary keys, and I felt that actor_id might have a higher cardinality than is preferred for primary keys. However, that's just an assumption on my part and I don't have benchmarks behind that one way or the other.

That said, I don't have a reason for leaving it out entirely and I think it should be added. I have a slight preference for including it in the ordering, between `enrollment_mode` and `emission_time`, but if you have a preference for the key then I'd say let's go with that.

bmtcril: I think it should be included since IIRC we're going to be grouping on it a lot to find the most recent state? I'm ok with it in that location. We can do some optimization tests later on the actual query patterns are established.

SoryRawyer: That makes sense. I'll include it in the other tables as well.

SoryRawyer: Ok, sorry, one last comment: I think you're right and that since the expected query pattern is to get the most recent state, `actor_id` should be ahead of `enrollment_mode`. In that case, I propose the following:
```sql
primary key (org, course_id)
order by (org, course_id, actor_id, enrollment_mode, emission_time)
```
What do you think?

bmtcril: Seems legit, let's try it!

bmtcril (approved):
👍

openedx-webhooks:
@SoryRawyer 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/tutor-contrib-oars&entry.867055334=https://github.com/openedx/tutor-contrib-oars/pull/82&entry.1484655318=https://github.com/SoryRawyer&entry.752974735=2023-06-22+16:04&entry.1917517419=2023-06-22+19:46&entry.2133058324=Yes) so we can improve your experience in the future.

PR #85, opened 2023-06-23
refactor!: Rename project from OARS to Aspects!

bmtcril:


Ian2012 (approved):
Probably some things will be broken, but it's better to do it now than block other PR by this work.

Some refactor is still pending for the docker images, the scripts and migrations and it's better to proceed with it

PR #88, opened 2023-06-23
refactor: Update oars-dbt repo to aspects-dbt

bmtcril:
Last project renaming PR, we've moved oars-dbt to aspects-dbt and need to update references here.


Ian2012 (approved):
LGTM

PR #141, opened 2023-07-05
feat!: Add ClickHouse RBAC

bmtcril:
WIP, the actual access controls don't work as expected currently and possibly conflict with the Superset RBAC (I'm reasonably certain we don't want both for performance reasons). This is just to collect feedback on the direction.


tutoraspects/templates/aspects/apps/superset/pythonpath/openedx_sso_security_manager.py
```diff
@@ -31,6 +32,10 @@ def add_bearer_jwt_token(token, uri, headers, body, placement="header"):
     return uri, headers, body
 
 
+def create_clickhouse_username(lms_username):
+    return f"openedx-{lms_username}"
```
bmtcril, line 36: For reference, this is the pattern Cairn uses. I opted not to change it, for backwards compat, but we still need to document the Cairn migration path and put in the required shims.

tutoraspects/templates/aspects/apps/superset/pythonpath/openedx_sso_security_manager.py
```diff
@@ -88,22 +101,73 @@ def access_token(self):
...
+            clickhouse_username, "openedx_instructor", "openedx" in user_roles
+        )
+        self._grant_or_revoke_clickhouse_role(
+            clickhouse_username, "openedx_superuser", "admin" in user_roles
+        )
+        self._grant_or_revoke_clickhouse_role(
+            clickhouse_username, "openedx_administrator", "alpha" in user_roles
+        )
+
+        make_clickhouse_query(f"SET DEFAULT ROLE ALL TO '{clickhouse_username}';")
```
bmtcril, line 126: This makes all of the above roles active for the user by default, necessary for them to be useful here.

tutoraspects/templates/aspects/apps/superset/pythonpath/superset_config.py
```diff
@@ -151,6 +151,19 @@ class CeleryConfig(object):
 BABEL_DEFAULT_LOCALE = "{{ SUPERSET_DEFAULT_LOCALE }}"
 LANGUAGES = {{ SUPERSET_SUPPORTED_LANGUAGES }}
 
+def DB_CONNECTION_MUTATOR(uri, params, username, security_manager, source):
```
bmtcril, line 154: This is where the actual magic happens. The ClickHouse username / password is replaced dynamically with the one we create for them at query time.

PR #191, opened 2023-07-19
Make Vector bulk import more idempotent

### LINKED ISSUE #127: Improve duplicate event handling in Vector
When using Vector, no checks are made to deduplicate event ids. On event replay especially this can be an issue, where a high quantity of duplicate events may land in a short period of time. A possible suggestion is to use [ReplacingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree) instead or MergeTree for the top level table type. More information is available [here](https://altinity.com/blog/clickhouse-replacingmergetree-explained-the-good-the-bad-and-the-ugly).

bmtcril:
This PR contains a new "tutor do" command to wrap bulk transform of tracking log events so that Vector can capture them, as well as various fixes to make the top level materialized view tables "eventually dedupllicated", which can help in certain edge cases with Ralph bulk import as well. 

Right now everything downstream of these tables is a view, and so will pick up deduplication automatically. If we need to start adding other downstream materializations we will need to take this into consideration and make sure to use the "FINAL" keyword or "OPTIMIZE FINAL" in their queries to prevent any duplicate rows trickling downstream.

Closes #127

### DIFF
--- a/tutoraspects/commands_v0.py
+++ b/tutoraspects/commands_v0.py
@@ -3,7 +3,7 @@ import click
 from tutor import config as tutor_config
 
 
-@click.command(help="Create an Open edX user and interactively set their password")
+@click.command(help="Run dbt with the provided command and options.")
 @click.option(
     "-c",
     "--command",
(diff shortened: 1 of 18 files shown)


bmtcril:
Heads up about these changes @SoryRawyer !

bmtcril:
I'm closing this for now, as I think #193 and #194 need to merge first and this will need a big refactor after those are in.

PR #193, opened 2023-07-24
refactor: parse course key in MV queries (FC-0024)

SoryRawyer:
Adds an alembic migration to update the materialized view queries. The course key is parsed from the `course_id` field in order to match the data in the `event_sink` tables.

This also updates all the Superset assets to use the new field. I don't think there are any required changes to the dbt models

### DIFF
--- a/tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0009_prefer_course_key.py
+++ b/tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0009_prefer_course_key.py
@@ -1,0 +1,260 @@
+from alembic import op
+
+revision = "0009"
+down_revision = "0008"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # n.b. we cann't rename the course_id column to course_key because it is used
+    # as part of the primary key for the MV target tables
+    op.execute("set allow_experimental_alter_materialized_view_structure=1;")
+    op.execute(
+        """
+        ALTER TABLE {{ ASPECTS_XAPI_DATABASE }}.{{ ASPECTS_ENROLLMENT_TRANSFORM_MV }}
+        MODIFY QUERY
+        SELECT
+            event_id,
+            emission_time,
+            actor_id,
+            object_id,
+            splitByString('/', course_id)[-1] as course_id,
+            org,
+            verb_id,
+            JSON_VALUE(event_str, '$.object.definition.extensions."https://w3id.org/xapi/acrossx/extensions/type"') AS enrollment_mode
+        FROM {{ ASPECTS_XAPI_DATABASE }}.{{ ASPECTS_XAPI_TABLE }}
+        WHERE verb_id IN (
+            'http://adlnet.gov/expapi/verbs/registered',
+            'http://id.tincanapi.com/verb/unregistered'
+        );
+    """
+    )
+
+    op.execute(
+        """
+        ALTER TABLE {{ ASPECTS_XAPI_DATABASE }}.{{ ASPECTS_VIDEO_PLAYBACK_TRANSFORM_MV }}
+        MODIFY QUERY
+        SELECT
+            event_id,
+            emission_time,
+            actor_id,
+            object_id,
+            splitByString('/', course_id)[-1] as course_id,
+            org,
+            verb_id,
+            cast(coalesce(
+                nullif(JSON_VALUE(event_str, '$.result.extensions."https://w3id.org/xapi/video/extensions/time"'), ''),
+                nullif(JSON_VALUE(event_str, '$.result.extensions."https://w3id.org/xapi/video/extensions/time-from"'), ''),
+                '0.0'
+            ) as Float32) as video_position
+        FROM {{ ASPECTS_XAPI_DATABASE }}.{{ ASPECTS_XAPI_TABLE }}
+        WHERE verb_id IN (
+            'http://adlnet.gov/expapi/verbs/completed',
+            'http://adlnet.gov/expapi/verbs/initialized',
+            'http://adlnet.gov/expapi/verbs/terminated',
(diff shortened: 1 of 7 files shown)


bmtcril (commented):
Overall this looks good, but I really don't want to carry the baggage of `course_id` forever. Given that we're still in development and no one has a huge chunk of data yet, I think what I'd like to see is dropping the MV target tables and reloading them using the query from the MV. You can probably move the queries out into their own variables and just reuse them. I know it's a pain but I think it will pay off in the long run.

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0009_prefer_course_key.py
```diff
@@ -0,0 +1,213 @@
...
+        """,
+}
+
+
+def upgrade():
+    # drop the tables and MVs and create new ones, with course_id
+    # replaced by course_key
+    for name, ddl in TABLES.items():
+        op.execute(f"DROP TABLE IF EXISTS {name}")
+        op.execute(ddl.replace("course_id", "course_key"))
```
bmtcril, line 195: Is there a reason not to just do this in the DDL?

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0009_prefer_course_key.py
```diff
@@ -0,0 +1,213 @@
...
+
+    for name, query in MATERIALIZED_VIEWS.items():
+        op.execute(f"DROP TABLE {name}")
+        op.execute(
+            query.replace(
+                "course_id", "splitByString('/', course_id)[-1] AS course_key"
+            )
+        )
+
```
bmtcril, line 204: We'll also want to populate the new tables with the results of the MV queries, sorry I wasn't clear what I was asking for there. I meant we could take the select statements and re-use them after create to do `insert into ... select from...`.

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0009_prefer_course_key.py
SoryRawyer: The only reason was to avoid having two copies of every query and DDL statement in the file, which I admit isn't really a great reason. I tried pulling the queries and DDL statements out into variables in the other migration files so I could import them here, but the Alembic file naming convention results in invalid module names (unless we did `importlib.import_module("0003_enrollment")`, which feels a little too awkward).

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0009_prefer_course_key.py
SoryRawyer: Ah got it, I'll change this so we create new tables, populate the data, then rename the existing tables before dropping them.

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0009_prefer_course_key.py
```diff
@@ -6,208 +8,430 @@
...
+
+def migrate(table_name, mv_name, table_ddl, mv_query):
+    # for each table and materialized view to migrate:
+    # - drop the existing table and materialized view if they exist
+    # - create the new table
+    # - load data into the new table using the new MV query
+    # - create the materialized view using the new query
+    op.execute(f"DROP TABLE IF EXISTS {table_name}")
+    op.execute(f"DROP TABLE IF EXISTS {mv_name}")
```
bmtcril, line 410: Does this actually do the drop of the mv? I would expect this to need a `drop view` but maybe fail silently if it doesn't find a table?

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0009_prefer_course_key.py
```diff
@@ -6,208 +8,430 @@
...
+CREATE TABLE IF NOT EXISTS {{ ASPECTS_XAPI_DATABASE }}.{{ ASPECTS_ENROLLMENT_EVENTS_TABLE }} (
+    `event_id` UUID NOT NULL,
+    `emission_time` DateTime64(6) NOT NULL,
+    `actor_id` String NOT NULL,
+    `object_id` String NOT NULL,
+    `course_id` String NOT NULL,
+    `org` String NOT NULL,
+    `verb_id` LowCardinality(String) NOT NULL,
+    `enrollment_mode` LowCardinality(String)
+) ENGINE = MergeTree
```
bmtcril, line 31: Last request, I think, but just to avoid having to rebuild all of these tables again for #191 can we make these tables `ReplacingMergeTree`? It will save a lot of complexity.

bmtcril (commented):
LGTM, but now that we've changed course_id to course_key I think we need to roll a new dbt version and include bumping that version in this update or it will break, right?

bmtcril:
There wasn't versioning happening on the dbt side, but I tagged the current master so people should be able to change their local settings temporarily if / when they run into issues. I'll post to the channel about it.

bmtcril (approved):

openedx-webhooks:
@SoryRawyer 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/tutor-contrib-aspects&entry.867055334=https://github.com/openedx/tutor-contrib-aspects/pull/193&entry.1484655318=https://github.com/SoryRawyer&entry.752974735=2023-07-24+13:32&entry.1917517419=2023-07-25+18:29&entry.2133058324=Yes) so we can improve your experience in the future.

PR #196, opened 2023-07-25
Fix migration when MV tables exist, upgrade Ralph

bmtcril:


SoryRawyer (approved):
Looks good! Apologies for the oversight

PR #205, opened 2023-07-26
Fixes for Vector bulk loading

bmtcril:
This PR contains a new "tutor do" command to wrap bulk transform of tracking log events so that Vector can capture them, as well as various fixes to make the top level materialized view tables "eventually dedupllicated", which can help in certain edge cases with Ralph bulk import as well.

Right now everything downstream of these tables is a view, and so will pick up deduplication automatically. If we need to start adding other downstream materializations we will need to take this into consideration and make sure to use the "FINAL" keyword or "OPTIMIZE FINAL" in their queries to prevent any duplicate rows trickling downstream.

Closes https://github.com/openedx/tutor-contrib-aspects/issues/127

### DIFF
--- a/tutoraspects/commands_v1.py
+++ b/tutoraspects/commands_v1.py
@@ -101,9 +101,22 @@ def dump_courses_to_clickhouse(options) -> list[tuple[str, str]]:
     return [("cms", f"./manage.py cms dump_courses_to_clickhouse {options}")]
 
 
+# pylint: disable=line-too-long
+# Ex: tutor local do transform-tracking-logs --options "--source_provider LOCAL --source_config '{\"key\": \"/openedx/data/\", \"prefix\": \"tracking.log\", \"container\": \"logs\"}' --destination_provider LRS --transformer_type xapi"
+# Ex: tutor local do transform-tracking-logs --options "--source_provider MINIO --source_config '{\"key\": \"openedx\", \"secret\": \"h3SIhXAqDDxJAP6TcXklNxro\", \"container\": \"openedx\", \"prefix\": \"/tracking_logs\", \"host\": \"files.local.overhang.io\", \"secure\": false}' --destination_provider LRS --transformer_type xapi"
+@click.command(context_settings={"ignore_unknown_options": True})
+@click.option("--options", default="", type=click.UNPROCESSED)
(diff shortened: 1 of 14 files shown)


tutoraspects/commands_v1.py
```diff
@@ -104,9 +104,22 @@ def dump_courses_to_clickhouse(options) -> list[tuple[str, str]]:
     return [("cms", f"./manage.py cms dump_courses_to_clickhouse {options}")]
 
 
+# pylint: disable=line-too-long
+# Ex: tutor local do transform-tracking-logs --options "--source_provider LOCAL --source_config '{\"key\": \"/openedx/data/\", \"prefix\": \"tracking.log\", \"container\": \"logs\"}' --destination_provider LRS --transformer_type xapi"
+# Ex: tutor local do transform-tracking-logs --options "--source_provider MINIO --source_config '{\"key\": \"openedx\", \"secret\": \"h3SIhXAqDDxJAP6TcXklNxro\", \"container\": \"openedx\", \"prefix\": \"/tracking_logs\", \"host\": \"files.local.overhang.io\", \"secure\": false}' --destination_provider LRS --transformer_type xapi"
+@click.command(context_settings={"ignore_unknown_options": True})
+@click.option("--options", default="", type=click.UNPROCESSED)
+def transform_tracking_logs(options) -> list[tuple[str, str]]:
```
Ian2012, line 112: This command is a bit hard to write, can you add inputs for every downstream input:

https://github.com/openedx/event-routing-backends/blob/dd4ce324b26021dac34859da2cb1b26b3a7ad27b/event_routing_backends/management/commands/transform_tracking_logs.py#L169-L232

bmtcril: I was really hoping to not have to do that, since it will mean having to keep them in sync with any changes to the upstream command, but I'd also really love to find a way to make this more usable. 🤔 🤔 🤔

Ian2012: What if we instead provide docs on how to do it using `tutor k8s exec lms ...`

bmtcril: I think based on our conversation in Slack we'll leave it for now, but follow up with #208 once we have Vector working in k8s.

Ian2012 (approved):

PR #206, opened 2023-07-26
feat!: reorder MVs and change data types (FC-0024)

SoryRawyer:
Updates the top-level materialized views in the following ways:
1. Adds fields to the `ORDER BY` clause to ensure the entire dataset is migrated during table updates
2. Truncates `emission_time` from `DateTime64(6)`, which stores nanosecond granularity, to `DateTime`, which stores second granularity
3. Stores video positions as `UInt32` (via `Decimal32(2)`) to avoid the imprecise `Float32` type

### DIFF
--- a/tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0011_reorder_mvs.py
+++ b/tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0011_reorder_mvs.py
@@ -1,0 +1,451 @@
+from dataclasses import dataclass
+
+from alembic import op
+
+revision = "0011"
+down_revision = "0010"
+branch_labels = None
+depends_on = None
+
+
+@dataclass
+class MvMigration:
+    table_name: str
+    old_ddl: str
+    new_ddl: str
+    mv_name: str
+    old_mv_query: str
+    new_mv_query: str
+
+
+OLD_ENROLLMENT_DDL = """
+CREATE OR REPLACE TABLE {{ ASPECTS_XAPI_DATABASE }}.{{ ASPECTS_ENROLLMENT_EVENTS_TABLE }} (
+    `event_id` UUID NOT NULL,
+    `emission_time` DateTime64(6) NOT NULL,
+    `actor_id` String NOT NULL,
+    `object_id` String NOT NULL,
+    `course_key` String NOT NULL,
+    `org` String NOT NULL,
+    `verb_id` LowCardinality(String) NOT NULL,
+    `enrollment_mode` LowCardinality(String)
+) ENGINE = ReplacingMergeTree
+PRIMARY KEY (org, course_key)
+ORDER BY (org, course_key, actor_id, enrollment_mode, emission_time);
+"""
+
+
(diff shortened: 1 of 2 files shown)


tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0011_reorder_mvs.py
```diff
@@ -0,0 +1,451 @@
+from dataclasses import dataclass
```
bmtcril, line 1: We should probably start commenting these as the large piles of SQL make it hard to see what changed. If you do a file level comment Alembic will print it out as part of the migration as well.

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0011_reorder_mvs.py
```diff
@@ -0,0 +1,451 @@
...
+
+from alembic import op
+
+revision = "0011"
```
bmtcril, line 5: I grabbed 11 and 12 (I think) in my PR, so this will need a rebase and update

bmtcril (commented):
Just a few comments, you should probably rebase on my PR that just went in to make sure everything still works as expected.

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0011_reorder_mvs.py
```diff
@@ -0,0 +1,451 @@
...
+    `emission_time` DateTime NOT NULL,
+    `actor_id` String NOT NULL,
+    `object_id` String NOT NULL,
+    `course_key` String NOT NULL,
+    `org` String NOT NULL,
+    `verb_id` LowCardinality(String) NOT NULL,
+    `enrollment_mode` LowCardinality(String)
+) ENGINE = ReplacingMergeTree
+PRIMARY KEY (org, course_key)
+ORDER BY (org, course_key, actor_id, emission_time, enrollment_mode, event_id);
```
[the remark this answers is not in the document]
SoryRawyer: Ah right, I remembered we wanted to put emission_time before the MV-specific fields, but forgot to move it ahead of actor_id. I don't have any benchmarks to suggest it performs better further down in the order by, just the more general ClickHouse suggestion that higher-cardinality fields should be further down the list of ordered fields. I think putting the emission_time right after the primary key fields makes sense.

bmtcril (approved):
👍 👍

bmtcril:
@SoryRawyer are there any downstream changes in the Superset assets that will be impacted by the type changes like we saw with the datetimes?

SoryRawyer:
@bmtcril not that I can see. The only data type change (aside from the timestamp change) is the `video_position` type, but that field isn't exposed in any datasets and the only time it's used in a virtual dataset query is after it's cast to an `Int32`. That cast shouldn't cause any issues, plus I have a change coming to that query anyway so it's a tiny bit of tech debt that'll get taken care of soon.

PR #214, opened 2023-07-27
fix: use DateTime instead of DateTime64 for datasets

bmtcril:
Some Superset assets needed to be updated to match the new DateTime columns


bmtcril:
@SoryRawyer heads up on this one, there may be more lingering issues like this in dbt as well

Ian2012 (approved):

SoryRawyer:
Ugh I'm sorry for missing these updates. I'll look at the dbt project and make sure everything's in order there.

PR #217, opened 2023-07-27
fix: specify insertion order in migration 0012 (FC-0024)

SoryRawyer:
ClickHouse materialized views insert data into the target table by referring to column names. However, when inserting data directly into the target table, it relies on column _order_ for the insertion. The query for the materialized views does not have the same ordering of org and course_id as the target table, and so when inserting data the order should be made explicit.


bmtcril (approved):
Thanks for this

bmtcril:
@Ian2012 heads up that this change is coming and you might have some data where the course and org are swapped because I'm dumb. We're just fixing this in the existing migration because we don't want real users to ever have those fields swapped.

PR #224, opened 2023-07-28
fix: deduplicate video timeline events (FC-0024)

SoryRawyer:
There was a missing join key that caused duplicate rows to be returned as part of the video timeline query. This resulted in inflated metrics for watched video segments.


bmtcril (approved):
I think this is happening in a number of places in the instructor charts, but will have to get a fresh start to confirm it since my local data is so messed up :)

SoryRawyer:
I was debating holding off on these fixes since I think these issues could be taken care of by moving to the new dictionary-backed tables, but I had a fix for this one already queued up.

PR #234, opened 2023-08-01
feat: use new dictionary-backed lookup tables to provide entity names (FC-0024)

SoryRawyer:
These changes should prevent the fan-out resulting from having multiple versions of a course or block and then joining to the event tables.

The datasets behind the dashboard filters were also updated to avoid relying on the `xblock_data_json` field. However, I left those tables pointing to the original `event_sink` tables since I'm not sure how to consistently parse all the different filter fields (which I think right now are: org, course name, run name, and block name).

### DIFF
--- a/tutoraspects/templates/openedx-assets/dim_courses.yaml
+++ b/tutoraspects/templates/openedx-assets/dim_courses.yaml
@@ -1,66 +1,54 @@
 - _file_name: dim_courses.yaml
   cache_timeout: null
   columns:
-  - advanced_data_type: null
-    column_name: course_key
-    description: null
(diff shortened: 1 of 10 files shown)


bmtcril (commented):
Are dim_course_problems / dim_course_videos still being used anywhere?

SoryRawyer:
They are; they provide the values for the dashboard filters. If the course key is guaranteed to always be part of a block's location, then we could probably rework those datasets to use the new dictionary-backed tables too.

bmtcril (approved):
👍  let's give it a try!

PR #238, opened 2023-08-01
feat: add transifex automatic translations

Ian2012:
## Description

This PR:

- Creates a workflow to automatically push strings to transifex.
- Creates a workflow to weekly update the translated strings.
- Updates the create_assets implementation to get strings from a new format grouped by lang


Ian2012:
Tests are failing just because the `locale.yaml` file is empty (to trigger the two translation workflows)

.github/workflows/compile-translations.yaml
```diff
@@ -0,0 +1,49 @@
...
+          git status
+          git log -l 3
+          git checkout -b bot/translations/${{ steps.date.outputs.date }}
+          git push -d origin bot/translations/${{ steps.date.outputs.date }}
+          git push origin bot/translations/${{ steps.date.outputs.date }}
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.EDX_TRANSIFEX_BOT_GITHUB_TOKEN }}
+          title: "chore(i18n): updating translations"
```
bmtcril, line 42: I don't think we support this style of conventional commit: https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0051-bp-conventional-commits.html#type should probably just be `chore:`?

tutoraspects/templates/aspects/apps/superset/pythonpath/create_assets.py
```diff
@@ -37,6 +39,7 @@
 
 ASSETS_TRANSLATIONS = yaml.load(open(TRANSLATIONS_FILE_PATH, "r"), Loader=yaml.FullLoader)
 
+print("ASSETS_TRANSLATIONS", ASSETS_TRANSLATIONS)
```
bmtcril, line 42: Do we still need this?

bmtcril (commented):
This is a really nice solution, just a couple little things

.github/workflows/compile-translations.yaml
Ian2012: I extract it from here https://github.com/openedx/edx-platform/pulls?q=is%3Apr+chore%28i18n%29+is%3Aclosed. So, should I use `chore:`

tutoraspects/templates/aspects/apps/superset/pythonpath/create_assets.py
Ian2012: nop

.github/workflows/compile-translations.yaml
bmtcril: Oh wild, I guess it will work. It's just not documented. 🤷 I think if you just remove that print statement it's good, then.

Ian2012:
@bmtcril ready

bmtcril (approved):

PR #239, opened 2023-08-01
docs: update virtual dataset links in README (FC-0024)

SoryRawyer:
The current links point to an old location of the Superset assets.


Ian2012:
@SoryRawyer could you use a `chore: update virutal dataset links in README (FC-0024)`

bmtcril:
I think technically it's "docs:"

PR #250, opened 2023-08-02
refactor: use dbt models for video datasets (FC-0024)

SoryRawyer:
This replaces the virtual datasets `fact_video_plays` and `fact_transcript_usage` with physical datasets that pull directly from dbt models. This PR relies on the changes in [this aspects-dbt PR](https://github.com/openedx/aspects-dbt/pull/19). If and when those changes are merged, I'll move this PR out of draft status.


bmtcril:
Just saw in the last PR that we're still not on versions for dbt, do you think you can squeeze that into this PR? Should just be able to change the branch [here](https://github.com/openedx/tutor-contrib-aspects/blob/main/tutoraspects/plugin.py#L205) to the next version that I'm about to release `v1.2`?

bmtcril (approved):

Issue #246, opened 2023-08-02
Bug: Instructor dashboard enrollments exclude current day


bmtcril:
Currently the default instructor dashboard filter is for the last month, but it ends that month at midnight of the current day:

```
SELECT toStartOfDay(toDateTime(`enrollment_status_date`)) AS `__timestamp`,
       count(*) AS `count`
FROM
...
WHERE `enrollment_status_date` >= '2023-08-01'
  AND `enrollment_status_date` < '2023-08-03'
  AND `org` IN ('UMONTREAL')
  AND `enrollment_status` = 'registered'
GROUP BY toStartOfDay(toDateTime(`enrollment_status_date`))
LIMIT 100000;
```

This is going to be confusing to instructors and cause a lot of issues on launch day of their course, when they most want to watch enrollments. We should find a way to make this inclusive in the same way that the other enrollment charts are.

SoryRawyer:
I think we can address this by using a "custom" time filter and setting the end of the window to one day after the current time. Does it make sense to leave the time frame start date as 1 month ago, or is there a better default value?

bmtcril:
That makes sense to me

PR #260, opened 2023-08-03
feat: include current day in instructor dashboard (FC-0024)

SoryRawyer:
This addresses #246. The query for `fact_enrollments_by_day` is also updated to extend the current enrollment status window to tomorrow so that the current days' enrollment status is included, which in turn makes the enrollment status available for the chart. In that same query, I've also confirmed that the call to `lagInFrame` provides a default value and so the call to `coalesce` further down is not needed.

### DIFF
--- a/tutoraspects/templates/openedx-assets/Instructor_dashboard_9.yaml
+++ b/tutoraspects/templates/openedx-assets/Instructor_dashboard_9.yaml
@@ -1,5 +1,5 @@
 - _file_name: Instructor_dashboard_9.yaml
-  _roles: 
+  _roles:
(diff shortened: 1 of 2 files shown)


SoryRawyer:
I realize now that the commit message/PR title uses `feat` but references an issue that starts with `bug`. Please let me know if it'd be better to amend the commit/PR title to use `fix` instead.

bmtcril (approved):
This is great, thanks. No worries about the commit message.

openedx-webhooks:
Although this pull request is already merged,  I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #261, opened 2023-08-03
fix: remove extra parens from dashboard filters (FC-0024)

### LINKED ISSUE #258: Bug: Instructor dashboard error when selecting more than one org
Looks like a datatype / extra parens issue:

```
:HTTPDriver for http://clickhouse:8123 returned response code 500)
 Code: 43. DB::Exception: Invalid types for IN function: String and Tuple(Tuple(String, String)).: While processing (verb_id = 'https://w3id.org/xapi/video/verbs/played') AND (org IN tuple(('edX', 'UMONTREAL'))). (ILLEGAL_TYPE_OF_ARGUMENT) (
This may be triggered by: 
Issue 1002 - The database returned an unexpected error.
```

SoryRawyer:
This fixes #258. The `where_in` jinja filter will add parentheses automatically. Since this macro was wrapped in parentheses in a few places, it caused an issue when more than one filter value was selected (though apparently it was fine when only one value was selected).

### DIFF
--- a/tutoraspects/templates/openedx-assets/queries/dim_course_problems.sql
+++ b/tutoraspects/templates/openedx-assets/queries/dim_course_problems.sql
@@ -12,10 +12,10 @@ with courses as (
         location like '%problem+block%'
     {% raw -%}
         {% if filter_values('org') != [] %}
-        and org in ({{ filter_values('org') | where_in }})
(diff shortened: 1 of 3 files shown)


bmtcril (approved):
\o/

PR #299, opened 2023-08-15
fix: add cache wrapper for can_view_courses filter

Ian2012:
This PR adds a cache wrapper for the `can_view_courses` function. The cache can be configured using the superset `data_cache`

It also removes the caching for the `get_courses` function such that people will get the latest courses at login.

### DIFF
--- a/tutoraspects/templates/aspects/apps/superset/pythonpath/openedx_sso_security_manager.py
+++ b/tutoraspects/templates/aspects/apps/superset/pythonpath/openedx_sso_security_manager.py
@@ -3,7 +3,7 @@ from collections import namedtuple
 
 import jwt
 from authlib.common.urls import add_params_to_qs, add_params_to_uri
-from flask import current_app, session, request
+from flask import current_app, session
 from superset.security import SupersetSecurityManager
 from superset.utils.memoized import memoized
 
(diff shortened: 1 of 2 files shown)


itsjeyd:
Thanks for these changes @Ian2012!

SoryRawyer (approved):
Looks good! I tried `memoized_func` locally and it seemed to work as expected so I think this should help cut down on dashboard loading times.

bmtcril (approved):
Seems to do the right thing, I'm not sure about removing the get_courses memo though. I think it would still get called repeatedly for different field names, which is likely to happen a few times in quick succession in a dashboard load. Current cache time for all caches if 5 mins, that seems reasonable enough for getting fresh permissions. Since it was monitoring the access key it would get refreshed on a new login as well. What do you think?

Ian2012:
We would have two caches, right?

Once for the `get_courses` and another one for the macro

bmtcril:
Right, so if the macro got called with "course id" and then again with "course key" theoretically we wouldn't have to hit the LMS twice.

openedx-webhooks:
@Ian2012 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/tutor-contrib-aspects&entry.867055334=https://github.com/openedx/tutor-contrib-aspects/pull/299&entry.1484655318=https://github.com/Ian2012&entry.752974735=2023-08-15+19:38&entry.1917517419=2023-08-25+18:10&entry.2133058324=Yes) so we can improve your experience in the future.

PR #308, opened 2023-08-18
fix: Changes needed to run in CH Cloud

bmtcril:
This fixes some issues I ran into with load testing in CH Cloud. One thing I haven't tested is how this all works on tutor local / dev so I'm putting in a PR for automation to check it. :)


Ian2012 (approved):

SoryRawyer (approved):
Thanks for making these changes! Looks good!

Issue #348, opened 2023-09-06
Bug: When running tutor initialization for this plugin I receive an error with port 9000 already in use.


ztraboo:
I noticed that I was receiving the following error when running `tutor dev init --limit` aspects.

**Error 0.0.0.0:9000: bind: address already in use**
```
  ⠋ Container tutor_nutmeg_dev-clickhouse-1  Starting                                                                                                                                                   0.1s
Error response from daemon: driver failed programming external connectivity on endpoint tutor_nutmeg_dev-clickhouse-1 (056f04d95199cef94e3547958160276a6a11e50a1dbddc9028e28a16ae81de28): listen tcp4 **0.0.0.0:9000: bind: address already in use**
Error: Command failed with status 1: docker-compose -f /Users/ztraboo/Dev/EducateWorkforce/Repos/open-edx/tutor.dev.nutmeg/tutor-env/env/local/docker-compose.yml -f /Users/ztraboo/Dev/EducateWorkforce/Repos/open-edx/tutor.dev.nutmeg/tutor-env/env/dev/docker-compose.yml -f /Users/ztraboo/Dev/EducateWorkforce/Repos/open-edx/tutor.dev.nutmeg/tutor-env/env/dev/docker-compose.tmp.yml -f /Users/ztraboo/Dev/EducateWorkforce/Repos/open-edx/tutor.dev.nutmeg/tutor-env/env/dev/docker-compose.override.yml --project-name tutor_nutmeg_dev -f /Users/ztraboo/Dev/EducateWorkforce/Repos/open-edx/tutor.dev.nutmeg/tutor-env/env/local/docker-compose.jobs.yml -f /Users/ztraboo/Dev/EducateWorkforce/Repos/open-edx/tutor.dev.nutmeg/tutor-env/env/dev/docker-compose.jobs.yml -f /Users/ztraboo/Dev/EducateWorkforce/Repos/open-edx/tutor.dev.nutmeg/tutor-env/env/dev/docker-compose.jobs.tmp.yml run --rm clickhouse-job sh -e -c echo “Initialising Clickhouse...”
```

After further investigation it appears that this port 9000 conflicts with my existing `tutor-minio` container. It looks like this port mapping is just for the tutor-dev environment. I'm working with nutmeg, however, I noticed that this minio port configuration was also present in v16.0.0 (palm)
[...]

PR #371, opened 2023-09-11
Update ClickHouse host port configuration

bmtcril:
The default host port conflicts with MinIO, this changes the default port and makes the host ports match the configured internal ports like we do for Supserset.

Closes: #348


bmtcril:
FWIW: I fully expect this to uncover bugs where that port is hard coded elsewhere, but it's best to find them now!

bmtcril:
This did uncover a bunch of small things, I've updated the port config names to reflect what they actually are (a number of native ports were named http) and successdfully tested locally in tutor local and dev with minio. I have not tested k8s, but I'm hoping CI will flush out any issues.

Ian2012:
@bmtcril can you perform a little test in Clickhouse cloud?

bmtcril:
Checked against CH Cloud, seems good. I think there's an unrelated (?) error publishing courses to CH Cloud from event-sink-clickhouse, though.

bmtcril:
I found and fixed the event sink error, it was unrelated but right on a line I touched earlier. 👎

Ian2012 (approved):

PR #398, opened 2023-09-18
feat: add materialized view for forum events (FC-0033)

SoryRawyer:
This adds a new top-level materialized view for forum events. This set of fields should cover our existing use cases so I didn't want to add anything else until we had a better idea of the requirements, but please let me know if there's anything else that could be included at this point.

### DIFF
--- a/tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0019_forum_events_mv.py
+++ b/tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0019_forum_events_mv.py
@@ -42,15 +42,7 @@ def upgrade():
         FROM
             {{ ASPECTS_XAPI_DATABASE }}.{{ ASPECTS_XAPI_TABLE }}
         WHERE
-            verb_id in (
-                'https://w3id.org/xapi/acrossx/verbs/posted',
-                'https://w3id.org/xapi/acrossx/verbs/edited',
-                'https://w3id.org/xapi/dod-isd/verbs/deleted',
-                'http://id.tincanapi.com/verb/viewed',
-                'https://w3id.org/xapi/openedx/verb/voted',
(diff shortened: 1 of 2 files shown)


tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0019_forum_events_mv.py
```diff
@@ -0,0 +1,64 @@
...
+            {{ ASPECTS_XAPI_DATABASE }}.{{ ASPECTS_XAPI_TABLE }}
+        WHERE
+            verb_id in (
+                'https://w3id.org/xapi/acrossx/verbs/posted',
+                'https://w3id.org/xapi/acrossx/verbs/edited',
+                'https://w3id.org/xapi/dod-isd/verbs/deleted',
+                'http://id.tincanapi.com/verb/viewed',
+                'https://w3id.org/xapi/openedx/verb/voted',
+                'https://w3id.org/xapi/acrossx/verbs/reported',
+                'https://w3id.org/xapi/openedx/verb/unreported'
```
Ian2012, line 52: @bmtcril what do you think about this selector?

I think other events will use the `deleted`, `edited`, `posted`, and `viewed` verbs

bmtcril: Hmm, maybe we can just filter these based on the object.definition.type being `http://id.tincanapi.com/activitytype/discussion` ?

Ian2012: That's better

Ian2012 (approved):

bmtcril (approved):

Issue #435, opened 2023-09-27
Build: Move translations to openedx Transifex


bmtcril:
We've been using a generously shared EduNEXT Transifex project so far for testing purposes, but need to move those translations into the openedx account so our translators can work on it!

PR #462, opened 2023-10-12
refactor: Move translations to the openedx Transifex project

bmtcril:
This tries to get us in line with OEP-58 and using openedx-translations / Atlas to manage the translation files. It's very likely going to require some iteration to get everything working correctly, but we can't do that until this lands since we need to get some actual translation files in Transifex to work with. So consider this part one of at least two.

Will need https://github.com/openedx/openedx-translations/pull/1834 to land to be functional as well, so I'm awaiting feedback on that before merging.

Partially solves: https://github.com/openedx/tutor-contrib-aspects/issues/435


Ian2012 (approved):

bmtcril:
@Ian2012 made a couple of changes based on feedback and to fix tests. Want to take another look?

OmarIthawi:
Amazing! Thanks for pushing on this front :)

Issue #406, opened 2023-09-21
Bug: Superset permissions cached indefinitely


bmtcril:
The security manager uses an LRU cache on LMS calls to get a user's courses, on a small site it may never expire leading to old permisisons such as not being able to log in to Superset or having instructor access when it has been revoked: https://github.com/openedx/tutor-contrib-aspects/blob/90a8afd161b4664352a3553b919a73a5ad8a4b98/tutoraspects/templates/aspects/apps/superset/pythonpath/openedx_sso_security_manager.py#L193C1-L194

We should either make this a time based cache or remove it altogether.

bmtcril:
We're seeing reports of this on the beta server again, even though I've confirmed it's got the fix for the LRU cache, so I'm going to reopen it. Waiting to hear if bouncing the server fixes it like it did last time. So far it's persisted over an hour without getting new permissions.

PR #489, opened 2023-10-24
Fixes for permissions cache and localization issues

bmtcril:
Fixes: #406

### DIFF
--- a/tutoraspects/templates/aspects/apps/superset/pythonpath/openedx_sso_security_manager.py
+++ b/tutoraspects/templates/aspects/apps/superset/pythonpath/openedx_sso_security_manager.py
@@ -6,6 +6,7 @@ import requests
 from authlib.common.urls import add_params_to_qs, add_params_to_uri
 from authlib.integrations.flask_client import OAuthError
 from flask import current_app, session
+from superset.extensions import cache_manager
 from superset.security import SupersetSecurityManager
 
 log = logging.getLogger(__name__)
@@ -78,6 +79,9 @@ class OpenEdxSsoSecurityManager(SupersetSecurityManager):
             user_roles = self._get_user_roles(
                 user_profile.get("preferred_username"), locale_preference
             )
+
+            log.info(f"User roles for {user_profile['preferred_username']}:"
+                     f" {user_roles}")
 
             return {
                 "name": user_profile["name"],
@@ -161,12 +165,15 @@ class OpenEdxSsoSecurityManager(SupersetSecurityManager):
         decoded_access_token = self.decoded_user_info()
 
         if decoded_access_token.get("superuser", False):
-            return ["admin", "admin-{locale}"]
+            return ["admin", f"admin-{locale}"]
         elif decoded_access_token.get("administrator", False):
-            return ["alpha", "operator", "operator-{locale}"]
+            return ["alpha", "operator", f"operator-{locale}"]
         else:
-            # User has to have staff access to one or more courses to view any content here.
-            courses = self.get_courses(username)
+            # User has to have staff access to one or more courses to view any content
+            # here. Since this is only called on login, we take the opportunity
+            # to force refresh the user's courses from LMS. This allows them to bust
+            # the course permissions cache if necessary by logging out and back in.
+            courses = self.get_courses(username, force=True)
             if courses:
                 return ["instructor", f"instructor-{locale}"]
             else:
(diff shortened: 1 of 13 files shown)


tutoraspects/templates/aspects/apps/superset/pythonpath/openedx_sso_security_manager.py
```diff
@@ -209,7 +230,9 @@ def get_courses(self, username, permission="staff", next_url=None):
             username=username, permission=permission
         )
         url = next_url or courses_url
-        response = oauth_remote.get(url, token=token).json()
+        resp = oauth_remote.get(url, token=token)
+        resp.raise_for_status()
```
bmtcril, line 234: Previously we ignored errors here which limited logging things like the LMS 429 rate limit errors, this will now raise them for easier debugging.

tutoraspects/templates/aspects/apps/superset/pythonpath/superset_config.py
```diff
@@ -69,8 +69,9 @@ def get_env_variable(var_name: str, default: Optional[str] = None) -> str:
 
 REDIS_HOST = get_env_variable("REDIS_HOST")
 REDIS_PORT = get_env_variable("REDIS_PORT")
-REDIS_CELERY_DB = get_env_variable("REDIS_CELERY_DB", "0")
-REDIS_RESULTS_DB = get_env_variable("REDIS_RESULTS_DB", "1")
+REDIS_CELERY_DB = get_env_variable("REDIS_CELERY_DB", "3")
+REDIS_RESULTS_DB = get_env_variable("REDIS_RESULTS_DB", "4")
+REDIS_CACHE_DB = get_env_variable("REDIS_CACHE_DB", "5")
```
bmtcril, line 74: Moving these to some other databases in redis so they are easier to identify. Previously they shared namespace with LMS which made debugging challenging and increased the possibility of name collisions.

tutoraspects/templates/aspects/apps/superset/pythonpath/superset_config.py
```diff
@@ -88,9 +89,9 @@ def get_env_variable(var_name: str, default: Optional[str] = None) -> str:
 
 # Cache for dashboard filter state
 FILTER_STATE_CACHE_CONFIG: CacheConfig = {
-    "CACHE_DEFAULT_TIMEOUT": int(timedelta(days=90).total_seconds()),
+    "CACHE_DEFAULT_TIMEOUT": 1200,
     # should the timeout be reset when retrieving a cached value
-    "REFRESH_TIMEOUT_ON_RETRIEVAL": True,
+    "REFRESH_TIMEOUT_ON_RETRIEVAL": False,
```
bmtcril, line 94: Bumping down the time and forcing these to refresh now, there could be some confusing artifacts when the filters got cached but a user's permissions had changed. This cuts down the possible length of time for that state to 20 mins. Artifacts were things like problem and video names appearing selected in the filters when the user no longer had access to them, though the charts do the right thing and do not display the data.

tutoraspects/templates/openedx-assets/assets/dashboards/Instructor_Dashboard.yaml
```diff
@@ -1,6 +1,6 @@
 _file_name: Instructor_Dashboard.yaml
 _roles:
-- {{ SUPERSET_ROLES_MAPPING.instructor }}
+- "{{ SUPERSET_ROLES_MAPPING.instructor }}"
 css: ''
```
bmtcril, line 4: For the localization process to work now we need to wrap all curly braces in double quotes. This still needs testing on the translations_pull side once the existing translations are updated, but I think it will work.

tutoraspects/templates/aspects/apps/superset/pythonpath/superset_config.py
Ian2012, line 74: Can we only use numbers? I think being explicit would be better

Ian2012 (approved):

tutoraspects/templates/aspects/apps/superset/pythonpath/superset_config.py
bmtcril: Explicit in what way?

bmtcril: Oh, like names for the redis dbs? No it doesn't support that, just creates dbs numbred 0-15 by default.

PR #516, opened 2023-11-08
Fix problem responses, bump ERB

bmtcril:


bmtcril:
@SoryRawyer this, combined with the upstream fix in ERB, should fix the problem responses issues we were seeing if you'd like to take a look.

Ian2012 (approved):

refresh_superset_assets.sh
Ian2012, line 1: what's this for?

bmtcril: Ah that was checked in by accident. It's how I deal with all the steps when working with assets these days.

PR #550, opened 2023-12-20
chore: Upgrade xapi-db-load to 1.0

bmtcril:
Upgrades xapi-db-load to the new config-based version along with the tests and commands that use it.


bmtcril:
Hmm, tests all passed but there are actually errors in the logs. >_>

```
clickhouse_connect.driver.exceptions.DatabaseError: :HTTPDriver for http://clickhouse:8123 returned response code 404)
 Code: 60. DB::Exception: Table xapi.xapi_events_all_parsed does not exist. Maybe you meant xapi_events_all?. 
```

bmtcril:
Oh, ha, dbt is creating everything in the reporting schema. That complicates things. cc @SoryRawyer 

`2 of 19 START sql materialized_view model `reporting`.`xapi_events_all_parsed` . [RUN]`

tutoraspects/templates/aspects/apps/aspects/scripts/xapi-db-load-config.yaml
Ian2012, line 1: @bmtcril can you add a patch here so that we can customize this file as needed?

tutoraspects/commands_v0.py
```diff
@@ -39,19 +39,18 @@ def dbt(context, command) -> None:
...
     """
     config = tutor_config.load(context.root)
     runner = context.job_runner(config)
 
-    command = f"""echo 'Making demo xapi script executable...'
-    echo 'Done. Running script...'
-    bash /app/aspects/scripts/clickhouse-demo-xapi-data.sh {num_batches} {batch_size}
+    command = f"""echo 'Running script...'
+    cd /app/aspects/scripts/
+    bash clickhouse-demo-xapi-data.sh {config_file}
```
Ian2012, line 53: The config_file parameter can't be used in docker/k8s unless we mounted it somehow in the aspects Pod. I think you should remove it and leave a patch or a setting that allows to modify the `xapi-db-load-config.yaml` and use that file by default

Ian2012 (changes requested):

bmtcril:
@Ian2012 this should be ready for re-review

Ian2012 (approved):

PR #554, opened 2024-01-04
refactor: Use named collections for dictionaries

bmtcril:
This allows us to change admin passwords and keep credentials out of the migrations.

Draft for now, until I have a chance to test against CH Cloud and Altinity, but I want to see what CI does.

### DIFF
--- a/tutoraspects/plugin.py
+++ b/tutoraspects/plugin.py
@@ -174,53 +174,6 @@ hooks.Filters.CONFIG_DEFAULTS.add_items(
             "{% endif %}",
         ),
         ("CLICKHOUSE_K8S_VOLUME_SIZE", "10Gi"),
-        # This can be used to override some configuration values in
-        # via "docker_config.xml" file, which will be read from a
-        # mount on /etc/clickhouse-server/config.d/ on startup.
-        # See https://clickhouse.com/docs/en/operations/configuration-files
-        #
-        # This default allows connecting to Clickhouse when run as a
-        # standalone docker container, instead of through docker-compose.
-        (
-            "CLICKHOUSE_EXTRA_XML_CONFIG",
-            """
-    <!-- Port for HTTP API. See also 'https_port' for secure connections.
-         This interface is also used by ODBC and JDBC drivers (DataGrip, Dbeaver, ...)
-         and by most of web interfaces (embedded UI, Grafana, Redash, ...).
-    -->
-    <http_port>{{CLICKHOUSE_INTERNAL_HTTP_PORT}}</http_port>
-
-    <!-- Port for interaction by native protocol with:
-         - clickhouse-client and other native ClickHouse tools (clickhouse-benchmark,
-           clickhouse-copier);
-         - clickhouse-server with other clickhouse-servers for distributed query processing;
-         - ClickHouse drivers and applications supporting native protocol
-         (this protocol is also informally called as "the TCP protocol");
-         See also 'tcp_port_secure' for secure connections.
-    -->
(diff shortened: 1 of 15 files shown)


tutoraspects/templates/aspects/jobs/init/clickhouse/init-clickhouse.sh
```diff
@@ -61,6 +61,30 @@ GRANT {{ ON_CLUSTER }} SELECT ON system.events TO '{{ ASPECTS_CLICKHOUSE_REPORT_
...
+DROP NAMED COLLECTION IF EXISTS local_ch_xapi;
+
+-- These are strictly for referencing local data, so localhost is appropriate
+CREATE NAMED COLLECTION local_ch_xapi AS
+host = 'localhost',
+port = '{{ CLICKHOUSE_INTERNAL_NATIVE_PORT }}',
+database = '{{ ASPECTS_XAPI_DATABASE }}',
+user = '{{ CLICKHOUSE_ADMIN_USER }}',
+password = '{{ CLICKHOUSE_ADMIN_PASSWORD }}',
+secure = {% if CLICKHOUSE_SECURE_CONNECTION %}1{% else %}0{% endif %};
```
Ian2012, line 76: does the secure flag makes sense in a local ch?

bmtcril: I'm not sure, that's what I'm going to be testing on the cloud providers.

tutoraspects/plugin.py
```diff
@@ -227,6 +227,38 @@
...
+        <local_ch_event_sink>
+            <host overridable="false">localhost</host>
+            <port overridable="false">{{ CLICKHOUSE_INTERNAL_NATIVE_PORT }}</port>
+            <database overridable="false">{{ ASPECTS_EVENT_SINK_DATABASE }}</database>
+            <user overridable="false">{{ ASPECTS_CLICKHOUSE_REPORT_USER }}</user>
+            <password overridable="false">{{ ASPECTS_CLICKHOUSE_REPORT_PASSWORD }}</password>
+            <secure overridable="false">{% if CLICKHOUSE_SECURE_CONNECTION %}1{% else %}0{% endif %}</secure>
+        </local_ch_event_sink>
+          """
+        ),
```
Ian2012, line 261: This is a too large to be a setting, can we use a patch for it and template it?

Also, as Clickhouse resolves any conflicts we can just override any value here using the patch

Ian2012 (commented):
Just a comment on the new setting

tutoraspects/plugin.py
bmtcril: I did this to match the other ClickHouse settings (`CLICKHOUSE_EXTRA_XML_CONFIG`, `CLICKHOUSE_EXTRA_USERS_XML_CONFIG`) so that they're not being configured in two different ways. I can make all of those patches, but I think that requires people to create a plugin to override them instead of just changing a config value? Just making sure since we would probably need to document the two types of settings clearly to say how to override the different settings.

bmtcril: It's also worth noting that `CLICKHOUSE_EXTRA_USERS_XML_CONFIG` in the current configuration doesn't work on k8s as well. I wasn't able to solve that problem as part of this.

Ian2012: That's a too large configuration to be managed as a setting. I would prefer the patch

bmtcril:
We can't actually use this until CH Cloud supports named collections / 23.11 build. So I'm closing this in favor of the work to get dictionaries into dbt, which would also solve this problem.

PR #572, opened 2024-01-18
Upgrade to Ralph 4.0

bmtcril:
Draft PR until Ralph 4 comes out of beta. Will need further cleanup and testing, as well as these prs to land first:

- https://github.com/openedx/aspects-dbt/pull/41
- https://github.com/openedx/xapi-db-load/pull/81
- https://github.com/openedx/tutor-contrib-aspects/pull/550


tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0029_drop_json_column.py
```diff
@@ -0,0 +1,91 @@
...
+
+def upgrade():
+    op.execute(
+        """
+        SET allow_experimental_object_type=1;
+        """
+    )
+    op.execute(
+        f"""
+        DROP VIEW IF EXISTS {{ ASPECTS_XAPI_DATABASE }}.xapi_events_all_parsed_mv;
```
bmtcril, line 26: We can't operate on the table as long as this view exists. It doesn't matter on a clean install where the mv doesn't exist yet, but if someone ever upgraded from an older version they would need to re-run dbt to get this back. Right now that's potentially a very expensive operation, but I guess they'd have to run dbt as part of the upgrade anyway. I'm not aware of any reason why people would downgrade / upgrade alembic outside of development.

bmtcril: This might be a good time to consider further squashing migrations and resetting the numbering, but I know @Ian2012 is already running some live services of this. 🤔

Ian2012: I would need to consider this with my team

bmtcril:
This PR is still a WIP, once the related dbt and xapi-db-load PRs land I'll pick it back up. It may require more work in the Superset datasets, I'm not sure yet. cc @SoryRawyer for context.

Ian2012 (approved):

PR #577, opened 2024-01-29
docs: Update readme to include building Superset image

bmtcril:
Without this step we no longer create the correct assets, and Superset is full of errors.


bmtcril:
@Ian2012 I think that we do not need to build the aspects image, but I can add that if I'm wrong.

Ian2012:
@bmtcril there is no need to build the Aspects image

Ian2012 (approved):

PR #606, opened 2024-02-16
fix: Fix Python 3.12 support, image building

bmtcril:
The Tutor standard is to replace pkg_resources with importlib_resources, so that's what I'm doing here. It should be backward compatible to Python 3.8, I've tested it with a fresh install on 3.8 and 3.12. We can't run CI with 3.12 yet, though, because not all Tutor packages support it yet, so this downgrades the version for image building.


tutoraspects/plugin.py
```diff
@@ -34,7 +34,7 @@
         ("RUN_RALPH", True),
         ("RUN_SUPERSET", True),
         ("DOCKER_IMAGE_ASPECTS", "edunext/aspects:{{ ASPECTS_VERSION }}"),
-        ("DOCKER_IMAGE_CLICKHOUSE", "clickhouse/clickhouse-server:23.8"),
+        ("DOCKER_IMAGE_CLICKHOUSE", "altinity/clickhouse-server:23.8.8.21.altinitystable"),
```
Ian2012, line 37: why are we switching to the altinity image?

tutoraspects/plugin.py
```diff
@@ -486,9 +486,8 @@
         )  # pylint: disable=no-member
 except AttributeError as e:
     for service, template_path, priority in MY_INIT_TASKS:
-        full_path: str = pkg_resources.resource_filename(
-            "tutoraspects", os.path.join("templates", *template_path)
-        )
+        templates_path = str(importlib_resources.files("tutoraspects") / "templates")
```
Ian2012, line 489: nit: can be defined outside of the for loop as `TEMPLATES_PATH`

tutoraspects/plugin.py
bmtcril: Bah,I was testing that when the error came in. Removed.

tutoraspects/plugin.py
bmtcril: I'm not worried about this one, I'd just like to get a build out

Ian2012 (approved):

Issue #546, opened 2023-12-18
Perf: Only recreate materialzed views when necessary


bmtcril:
With the move of MVs to dbt, we currently rebuild every mv table on every dbt run / aspects init. This will be a problem with large datasets over time, we want to only rebuild those tables when absolutely necessary. dbt has the `on_configuration_change` configuration option to support this, we'll need to upstream it to dbt-clickhouse then update. 

See: https://docs.getdbt.com/docs/build/materializations#materialized-view

Ian2012:
@bmtcril are new MVs backfilled on creation?

bmtcril:
@Ian2012 yes, and on every init / dbt run currently

SoryRawyer:
After looking into this, the `on_configuration_change` setting is unlikely to prevent us from rebuilding the MVs on every run. As I understand it, there are two main approaches for us to avoid rebuilding MVs when we don't need to:
- The [`state:modified` selector](https://docs.getdbt.com/reference/node-selection/methods#the-state-method), which would require us to track the state of each dbt run over time, since that selector requires the previous runs' manifest to compare to.
- Adding a tag to each MV and selecting all models that don't have this tag. One approach could be to only select models materialized as views via [the `config` selector](https://docs.getdbt.com/reference/node-selection/methods#the-config-method)

PR #628, opened 2024-02-28
feat: Use dbt state to avoid large table rebuilds

bmtcril:
This is a work in progress to avoid rebuilding our large materialized views when we can avoid it. This dbt feature is complicated and requires storing the state of the last run so it can decide what's changed. 

Closes: https://github.com/openedx/tutor-contrib-aspects/issues/546

Open questions:

- How to work around this option being automatically added for commands that it doesn't make sense for (maybe just document that it needs to be turned off for them?)
- Is a k8s PVC the right way to go to store the state there?
- Are there other ways to surface that this is happening, or is some documentation enough
- What things should we consider safe to ignore for "changed" state? Tests, probably. Anything else? https://docs.getdbt.com/reference/node-selection/state-comparison-caveats

### DIFF
--- a/tutoraspects/templates/aspects/apps/aspects/scripts/dbt.sh
+++ b/tutoraspects/templates/aspects/apps/aspects/scripts/dbt.sh
@@ -26,11 +26,28 @@ cd aspects-dbt || exit
 
 export ASPECTS_EVENT_SINK_DATABASE={{ASPECTS_EVENT_SINK_DATABASE}}
 export ASPECTS_XAPI_DATABASE={{ASPECTS_XAPI_DATABASE}}
+export DBT_STATE={{ DBT_STATE_DIR }}
 
 echo "Installing dbt dependencies"
 dbt deps --profiles-dir /app/aspects/dbt/
 
-echo "Running dbt $*"
-dbt "$@" --profiles-dir /app/aspects/dbt/
+echo "Running dbt ${@:2}"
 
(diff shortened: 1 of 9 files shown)


Ian2012 (approved):

tutoraspects/templates/aspects/apps/aspects/scripts/dbt.sh
```diff
@@ -26,11 +26,28 @@ cd aspects-dbt || exit
...
+if [ "$1" == "True" ]
+then
+  echo "Requested to only run modified state, checking for ${DBT_STATE}/manifest.json"
+fi
+
+# If state exists and we've asked to only run changed files, add the flag
+if [ "$1" == "True" ] && [ -e "${DBT_STATE}/manifest.json" ]
+then
+  echo "Found {{DBT_STATE_DIR}}/manifest.json so only running modified items and their downstreams"
+  dbt "${@:2}" --profiles-dir /app/aspects/dbt/ -s state:modified+
```
SoryRawyer, line 45: The dbt docs aren't clear on this, but I think we also need to specify where dbt should look for the old manifest by passing `--state ${DBT_STATE}` to the dbt command.

bmtcril: The `DBT_STATE` env var works too, so I'm using that above to keep the command from getting ridiculous

Issue #626, opened 2024-02-27
Bug: Some dashboard types don't have a "name" attribute


bmtcril:
When importing the mockup dashboard I ran into the following:

```
  File "/app/pythonpath/create_assets.py", line 231, in generate_translated_dashboard_filters
    translation = get_translation(filter["name"], language)
```

The issue is that DIVIDER types don't have a name, but they do have a description we could key translations off of so we should check for both.

PR #629, opened 2024-02-28
fix: Some dashboard assets do not have a name to localize

bmtcril:
Some assets have a description instead of a name, so we now check that as well and fail with a clearer error if neither exists.

Closes: https://github.com/openedx/tutor-contrib-aspects/issues/626

### DIFF
--- a/tutoraspects/templates/aspects/apps/superset/pythonpath/create_assets.py
+++ b/tutoraspects/templates/aspects/apps/superset/pythonpath/create_assets.py
@@ -227,9 +227,9 @@ def generate_translated_dashboard_filters(copy, language):
     metadata = copy.get("metadata", {})
 
     for filter in metadata.get("native_filter_configuration", []):
-        translation = get_translation(filter["name"], language)
-
-        filter["name"] = translation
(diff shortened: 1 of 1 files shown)


tutoraspects/templates/aspects/apps/superset/pythonpath/create_assets.py
```diff
@@ -227,9 +227,15 @@ def generate_translated_dashboard_filters(copy, language):
     metadata = copy.get("metadata", {})
 
     for filter in metadata.get("native_filter_configuration", []):
-        translation = get_translation(filter["name"], language)
-
-        filter["name"] = translation
+        if "name" in filter:
+            translation = get_translation(filter["name"], language)
+            filter["name"] = translation
+        elif "description" in filter:
```
Ian2012, line 233: What if a filter has both? Can you separate this?

bmtcril: All set, it will now translate either if they exist and not error if those keys don't exist.

Ian2012 (approved):

PR #714, opened 2024-04-11
fix: Handle clickhouse errors in performance metrics

bmtcril:
Needed these changes to handle cases where queries got killed by ClickHouse for memory issues.


tutoraspects/templates/aspects/apps/superset/pythonpath/performance_metrics.py
```diff
@@ -87,7 +89,11 @@ def measure_chart(slice, extra_filters=[]):
     command = ChartDataCommand(query_context)
 
     start_time = datetime.now()
-    result = command.run()
+    try:
+        result = command.run()
+    except ChartDataQueryFailedError as e:
+        print(e)
+        result = {"result": "error", "data": None, "queries": []}
```
Ian2012, line 96: Please rebase on main, this error is already handled, but not as this

PR #737, opened 2024-04-22
fix: Remove query context after asset import

bmtcril:
Fixes an issue where dashboards do not load in the LMS instructor dashboard due to query_context containing database primary keys from the database it was exported from. Performance metrics uses the query_context, so now it pulls that directly from the asset files on disk instead.

### DIFF
--- a/tutoraspects/templates/aspects/apps/superset/pythonpath/performance_metrics.py
+++ b/tutoraspects/templates/aspects/apps/superset/pythonpath/performance_metrics.py
@@ -1,12 +1,17 @@
-import sys
+"""
+Gather performance metrics on Superset chart queries.
 
-from superset.app import create_app
+Reads the queries from the superset database, and enriches them with the
+query_context from the asset files. The query_context cannot be stored in the
+database on import due to is using database primary keys which do not match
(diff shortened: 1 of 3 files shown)


tutoraspects/templates/aspects/apps/superset/pythonpath/performance_metrics.py
```diff
@@ -78,16 +87,33 @@ def performance_metrics(course_key, print_sql):
...
+                continue
+
+            path = os.path.join(root, file)
+            with open(path, "r") as file:
+                asset = yaml.safe_load(file)
+                if "query_context" in asset and asset["query_context"]:
+                    query_contexts[asset["uuid"]] = json.loads(asset["query_context"])
+
+    logger.info(f"Found {len(query_contexts)} query contexts")
+    return query_contexts
```
Ian2012, line 109: what about returning a dictionary with all the context matched against the UUID. Then use it to access it directly below

Ian2012 (approved):
LGTM, only one comment for reading assets files once

Ian2012 (approved):

PR #753, opened 2024-04-29
fix: Fail Tutor command on performance_metrics failure

bmtcril:
I expect CI on this will fail until #750 is merged and I can rebase. I just want to make sure this fails SQL errors in the future.


Ian2012 (approved):

bmtcril:
@Ian2012 you may want to look at the last 2 commits here. They're fixes for the grade distribution charts.

Ian2012 (approved):

PR #757, opened 2024-04-30
fix: Normalize localized asset UUIDs

bmtcril:
This fixes mismatches between this plugin and the assets ids generated in the LMS plugin that cause translated dashboards to not embed properly. There is a matching PR in platform-plugin-aspects that is needed to complete this fix.


bmtcril:
This is going to fail until platform-plugin-aspects 0.7.3 releases.

Ian2012 (approved):

bmtcril:
Closing and reopening to run tests

PR #763, opened 2024-05-03
Upgrade aspects-dbt to 3.22

bmtcril:
Adds support for new block ordering and enrollment timestamp to at risk learners model.


SoryRawyer (approved):
looks good, thanks!

Ian2012 (approved):

PR #767, opened 2024-05-03
Altinity / clustering fixes

bmtcril:
Fixes to running ClickHouse in a clustered mode, such as how it is run in Altinity and in the k8s helm chart. Especially:

- Adds a new setting to force DDL operations (including dbt) to run on a single node so that interdependencies will always exist
- Forces drops to complete synchronously so that versions of tables/views don't get out of sync across replicas
- Remove "on cluster" from function creation since that no longer works- functions are automatically created on all nodes in the cluster


Ian2012:
@bmtcril are functions created on cluster automatically?

bmtcril:
@Ian2012 Functions created "on cluster" may be ClickHouse version dependent, but I believe I did test this with 24.3 (Aspects current version). I'll double check it tomorrow. I do know that I was getting errors that you can't specify "on cluster" for functions and needed this change.

Ian2012:
It needs the on cluster clause: https://clickhouse.com/docs/en/sql-reference/statements/create/function

This was tested on a clickhouse k8s cluster with two machines: 

DDL Node:
```
SELECT name
FROM system.functions
WHERE origin = 'SQLUserDefined'

Query id: 3acce4f2-7593-4a69-8b62-d68e6a8254d4

   ┌─name────────────────────────┐
1. │ get_org_from_ccx_course_url │
2. │ get_org_from_course_url     │
   └─────────────────────────────┘
```

Replica:
```
SELECT name
FROM system.functions
WHERE origin = 'SQLUserDefined'

Query id: a719e59d-1edb-4b20-8b36-8c5b4b944fe8

Ok.
```

bmtcril:
@Ian2012 which version of ClickHouse is that?

Ian2012:
24.3

bmtcril:
I think that with help from Altinity we've found the problem. On the Altinity operator we need to set a setting:

```
<clickhouse>
    <user_defined_zookeeper_path>/udf</user_defined_zookeeper_path>
</clickhouse>
```

This will allow the UDFs to propagate to all nodes as the other supported install methods do. We shouldn't need `ON CLUSTER` for create functions anywhere. 

We found a similar issue that might impact users / permissions: https://github.com/Altinity/clickhouse-operator/issues/886 that we'll look at separately.

Ian2012:
Can you add that to the clickhouse server settings? For reference

bmtcril:
@Ian2012 do you mean the docs, or the actual settings? I'm not sure what that setting will do in a non-keeper environment, and am not in a good state to find out.

Ian2012:
For the docs then

tutoraspects/commands_v1.py
```diff
@@ -27,7 +27,7 @@ def load_xapi_test_data(config_file) -> list[tuple[str, str]]:
     return [
         (
             "aspects",
-            "echo 'Running script...' && "
+            f"echo 'Running script... {config_file}' && "
```
bmtcril, line 30: The xapi-db-load changes are unrelated, but are useful for debugging what's actually being run there.

bmtcril:
@Ian2012 @saraburns1 this is ready for review when you have a chance

saraburns1 (approved):
Ran alembic downgrade to base and upgrade to head - no issues

Ian2012:
@bmtcril should we add integration tests for this?

bmtcril:
@Ian2012 what did you have in mind for integration tests?

Ian2012:
I was thinking on running a k8s clickhouse installation with the operator and test against that, but we already have 9 CI checks

bmtcril:
@Ian2012 yeah I think that would probably be more than we could do on an Action anyway. I'll just keep it as a manual release test step for now.

Ian2012 (approved):

PR #773, opened 2024-05-07
fix: Import on openedx-translations

bmtcril:
The path on the openedx-translations runner for extract-translations is different, and breaks with this new import. I'm changing it back to what it was before the script was moved. It works locally.


bmtcril:
Tested in openedx-translations successfully

Ian2012 (approved):

PR #810, opened 2024-05-13
build: don't tag this in Open edX releases

kdmccormick:
rationale -- So far we haven't tagged any Tutor plugins as part of a named release, as plugins are considered consumers of the release rather than components of the release itself.

___________________________

@bmtcril let's chat in standup before merging this. I was thinking more about this, I want to make sure you have a good way of marking which Aspects version works with which Open edX release. If you're not going to use version numbers like other Tutor plugins, then it might actually make sense to keep doing the release tags here. I have the same thought about aspects-dbt.


kdmccormick:
related details: https://github.com/openedx/tutor-contrib-aspects/pull/813

bmtcril:
@kdmccormick ok with you if I take this over and get it over the line?

kdmccormick:
@bmtcril All you 👍🏻

bmtcril (approved):

Issue #793, opened 2024-05-09
Bug: % correct and % incorrect problem attempts showing up as decimals


crathbun428:
Steps to Reproduce
1. Navigate to the Engagement/Problems tab on the Course Dashboard, Individual Learner Reports, and At-Risk Learner Reports
2. View the % correct and % incorrect columns in the Problem Results table

Expected Results
1. The data in these columns show up as percentages

Actual Results
1. The data in these columns show up as decimals, see below:
![Screenshot 2024-05-09 at 3 59 08 PM](https://github.com/openedx/tutor-contrib-aspects/assets/57507106/aeadc02e-a842-40b7-a8c2-17a97de556a1)

crathbun428:
@Ian2012 - This appears to be done, correct? If so - I can close it out. 

cc: @bmtcril

bmtcril:
The last fixes should be in #822 , so I can close this when that merges

Issue #792, opened 2024-05-09
Bug: Help tab is not present in instructor facing dashboards


bmtcril:
We need to re-implement the help tab for course / at-risk / individual learner dashboards

PR #822, opened 2024-05-22
At-risk dashboard performance enhancements

bmtcril:
This improves performance of the various charts of the at-risk dashboard. In the 300M row dataset all charts load now, though some are still slow. The problem engagement charts should be able to be improved by other changes in flight.

This also bumps platform-plugin-aspects to 0.9.4 and adds its new setting

Closes: #793 
Closes: #792


tutoraspects/patches/openedx-common-settings
```diff
@@ -15,6 +15,7 @@ EVENT_SINK_CLICKHOUSE_PII_MODELS = {{ EVENT_SINK_PII_MODELS }}
 
 ASPECTS_INSTRUCTOR_DASHBOARDS = {{ ASPECTS_INSTRUCTOR_DASHBOARDS }}
 SUPERSET_DASHBOARD_LOCALES = {{ SUPERSET_DASHBOARD_LOCALES }}
+SUPERSET_SHOW_INSTRUCTOR_DASHBOARD_LINK = {{ SUPERSET_SHOW_INSTRUCTOR_DASHBOARD_LINK }}
```
bmtcril, line 18: Needed for platform-plugin-aspects 0.9.4

tutoraspects/plugin.py
```diff
@@ -346,7 +348,7 @@
         # For now we are pulling this from github, which should allow maximum
         # flexibility for forking, running branches, specific versions, etc.
         ("DBT_REPOSITORY", "https://github.com/openedx/aspects-dbt"),
-        ("DBT_BRANCH", "v3.24.0"),
+        ("DBT_BRANCH", "bmtcril/performance_tweaks"),
```
bmtcril, line 351: Need to update this once https://github.com/openedx/aspects-dbt/pull/94 lands

tutoraspects/plugin.py
```diff
@@ -86,6 +86,8 @@
             "* [Superset Resources](https://github.com/apache/superset#resources)<br>",
         ),
         ("ASPECTS_ENABLE_INSTRUCTOR_DASHBOARD_PLUGIN", True),
+        # Whether to show the link to go to Superset in the instructor dashboard tab
+        ("SUPERSET_SHOW_INSTRUCTOR_DASHBOARD_LINK", True),
```
Ian2012, line 90: Do we need this setting? Operators can do the same using an inline plugin such as:

```yaml
name: aspects-settings
version: 0.1.0
patches:
    openedx-common-settings: |
      SUPERSET_SHOW_INSTRUCTOR_DASHBOARD_LINK = True
      # More aspects settings
```

tutoraspects/templates/aspects/build/aspects-superset/openedx-assets/assets/charts/At-risk_learners.yaml
```diff
@@ -13,18 +13,55 @@ params:
   - name
   - username
   - email
+  - expressionType: SQL
+    label: Course Grade %
+    sqlExpression: round(course_grade*100, 2)
```
Ian2012, line 18: This one should be updated on DBT, it would simplify translations.

tutoraspects/templates/aspects/build/aspects-superset/openedx-assets/assets/dashboards/At-Risk_Learners.yaml
```diff
@@ -1038,6 +1077,19 @@ position:
...
+    id: TAB-1sC34gQvt
+    meta:
+      defaultText: Tab title
+      placeholder: Tab title
+      text: Help
+    parents:
+    - ROOT_ID
+    - GRID_ID
+    - TABS-7McownEo06
+    type: TAB
```
Ian2012, line 1092: Remember to add the help markdown text here.

tutoraspects/templates/openedx-assets/queries/at_risk_learner_filter.sql
```diff
@@ -0,0 +1,18 @@
+with
+    page_visits as (
+        select org, course_key, actor_id, max(last_visited)
+        from {{ ASPECTS_XAPI_DATABASE }}.fact_learner_last_course_visit
+        where
+            1=1
+            {% include 'openedx-assets/queries/common_filters.sql' %}
+            and last_visited < subtractDays(now(), 7)
+        group by org, course_key, actor_id
+    )
```
Ian2012, line 10: Data is already grouped/merged on the downstream MV. Also, as the table is partitioned per org/course_key you don't need to add the common filters.

tutoraspects/templates/openedx-assets/queries/dim_at_risk_learners.sql
```diff
@@ -0,0 +1,30 @@
+with
+    page_visits as (
+        select org, course_key, actor_id, max(last_visited) as last_visited
+        from {{ ASPECTS_XAPI_DATABASE }}.fact_learner_last_course_visit
+        where
+            1=1
+            {% include 'openedx-assets/queries/common_filters.sql' %}
+        group by org, course_key, actor_id
+    )
```
Ian2012, line 9: Same as above, usually the MV with org/course_key partition is good enough to improve the query times

Ian2012 (commented):
A couple of comments for the sql changes.

tutoraspects/plugin.py
bmtcril: It would be weird that we set a bunch of settings for that plugin here, but not this one. I think it makes more sense to have there be defaults from the same place rather than have some missing.

tutoraspects/templates/aspects/build/aspects-superset/openedx-assets/assets/charts/At-risk_learners.yaml
bmtcril: Can you explain that? We do the `round()` thing in a lot of places

tutoraspects/templates/openedx-assets/queries/at_risk_learner_filter.sql
bmtcril: The grouping isn't entirely true, we still need it to narrow down unmerged blocks: 

> Data deduplication occurs only during a merge. Merging occurs in the background at an unknown time, so you can’t plan for it. Some of the data may remain unprocessed. Although you can run an unscheduled merge using the OPTIMIZE query, do not count on using it, because the OPTIMIZE query will read and write a large amount of data.
> 
> Thus, ReplacingMergeTree is suitable for clearing out duplicate data in the background in order to save space, but it does not guarantee the absence of duplicates.

The common filters is what makes this more performant (at least until the new CH analyzer lands in CH Cloud), otherwise it will pull all rows here and filter them in the join step below.

tutoraspects/templates/aspects/build/aspects-superset/openedx-assets/assets/dashboards/At-Risk_Learners.yaml
bmtcril: It's actually added at line 875, this is just the tab label itself. This whole yaml model is super confusing. :/

SoryRawyer (commented):
All of the SQL changes look good to me!

bmtcril:
Closing/reopening to re run tests with latest dbt changes

tutoraspects/templates/openedx-assets/queries/at_risk_learner_filter.sql
Ian2012: `last_visited` needs to be renamed to `emission_time` in the select statement

PR #823, opened 2024-05-23
fix: Use enrolled date if last visited is not set

saraburns1:
Change SQL to user greater of enrolled at date or last visited - 'last visited' defaults to 1970/1/1 on creation.

Closes: https://github.com/openedx/platform-plugin-aspects/issues/55


bmtcril:
CLA check is incoming, should pass later today!

Ian2012 (approved):

Issue #788, opened 2024-05-09
Operator dash shows incorrect user information with PII off


bmtcril:
When `ASPECTS_ENABLE_PII` is False the models that the Operator dashboard uses for the `Total Users` chart aren't populated. We need to use a different model for that.

PR #839, opened 2024-05-29
fix: Operator dash user counts, performance

bmtcril:
Don't use the user_pii table for user counts, since it may not be populated. Change this chart to be a count of active enrollments, which people have said is more useful.

Also changed the time range for operator dash to 90 days to improve performance of the actors / events graphs.

Closes: #788


tutoraspects/templates/openedx-assets/queries/fact_enrollment_pii.sql
```diff
@@ -8,7 +8,10 @@ with
         from {{ DBT_PROFILE_TARGET_DATABASE }}.fact_enrollment_status enrollment
         inner join
             {{ ASPECTS_EVENT_SINK_DATABASE }}.course_names as course_names
-            on course_names.course_key = enrollment.course_key
+        -- Need to cast the course key to a string here otherwise the
+        -- course_names dictionary throws this:
+        -- Key type for complex key at position 0 does not match, expected String, found LowCardinality(String).
+        on course_names.course_key = enrollment.course_key::String
```
Ian2012, line 14: Doesn't this affect DBT reports too?

tutoraspects/plugin.py
```diff
@@ -355,7 +355,7 @@
         # For now we are pulling this from github, which should allow maximum
         # flexibility for forking, running branches, specific versions, etc.
         ("DBT_REPOSITORY", "https://github.com/openedx/aspects-dbt"),
-        ("DBT_BRANCH", "v3.28.1"),
+        ("DBT_BRANCH", "bmtcril/update_instance_mvs"),
```
bmtcril, line 358: Needs to be changed before merge: https://github.com/openedx/aspects-dbt/pull/98

tutoraspects/templates/openedx-assets/queries/fact_enrollment_pii.sql
Ian2012: And custom reports that do the same join such as the `fact_*_engagement` ones?

bmtcril: The `fact_*_engagement` charts are working as-is, but I can't guarantee what other custom reports might run into. That's likely a problem with a number of changes we've been making, though.

Ian2012 (approved):

PR #842, opened 2024-05-30
fix: new strings for translation

### LINKED ISSUE #840: Extract additional dashboard strings for translation
Currently we are missing some strings from Dashboard translations:

- `chart.meta.sliceNameOverride`

We should be able to test this by running `make extract_translations` to debug and make sure the strings get picked up correctly.

saraburns1:
Closes #840 

Add 2 new strings for translation
Reconfigure method for easier reading
Fix 'could not translate' errors

New strings:
image

Logs before:
image

Logs After:
image

### DIFF
--- a/tutoraspects/translations/translate_utils.py
+++ b/tutoraspects/translations/translate_utils.py
@@ -45,14 +45,14 @@ class TranslatableAsset:
         """
         if not content or isinstance(content, str):
             return []
-        
+
         # if content is a list, run method for each list item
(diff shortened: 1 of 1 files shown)


Ian2012 (approved):

bmtcril (approved):
Ran it locally and it looks good! The only unexpected change in the output is:

`  '': ''`

Which I think is fine. Should be the easiest translation for the team to review. :)

PR #851, opened 2024-06-12
feat: Add command to generate a dbt exposures file

bmtcril:
We'll use this in aspects-dbt CI to keep the model to dashboard mappings up to date. Leaving this in draft for the moment so we can discuss the direction. The idea is to have this command to generate an exposures file that we can bring back to aspects-dbt.

This should allow us visibility into which datasets are actually being used (or not used) in Superset and let us remove unused models. However there's a dependency problem where tutor-contrib-aspects needs to update to use a new aspects-dbt and there may be error-causing issues generating the exposures until it is updated. I haven't given much thought to how to resolve that yet, but we could move the dbt docs to that repo instead and use the combined official tutor-contrib-aspects / aspects-dbt version to make the file / docs.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -76,6 +76,8 @@ jobs:
           tutor local do dump-data-to-clickhouse --options "--object course_overviews --force"
           make extract_translations
           tutor local do import-assets
+          tutor local do collect-dbt-lineage
+          tutor local run lms python manage.py lms transform_tracking_logs --source_provider LOCAL --source_config '{"key": "/openedx/data/", "prefix": "tracking.log", "container": "logs"}' --destination_provider LRS --transformer_type xapi
       - name: Performance metrics
         run: tutor local do performance-metrics --fail_on_error
       - name: Tutor stop
--- a/tutoraspects/templates/aspects/apps/superset/pythonpath/collect_dbt_lineage.py
+++ b/tutoraspects/templates/aspects/apps/superset/pythonpath/collect_dbt_lineage.py
@@ -1,0 +1,195 @@
+"""
+Parse all Superset datasets for SQL, find links back to dbt assets.
+
+This is used in aspects-dbt to update the exposures.
+"""
+
+# Necessary to access Superset
+from create_assets import app
+
+import json
+import logging
+import os
+import re
+
+import click
+import ruamel.yaml
+from superset.extensions import db
+from superset.models.dashboard import Dashboard
+
+
+logger = logging.getLogger("collect_dbt_lineage")
+
+DBT_DATA_PATH = "/app/dbt_manifest/"
+DBT_MANIFEST_PATH = os.path.join(DBT_DATA_PATH, "manifest.json")
+
+
+def get_tables_from_dbt():
+    """
+    Take generated metadata from the last dbt run to find known models
+    """
+    with open(DBT_MANIFEST_PATH) as f:
+        dbt_manifest = json.load(f)
+
+    tables = {}
+    for table_type in ["nodes", "sources"]:
+        manifest_subset = dbt_manifest[table_type]
+
+        for table_key_long in manifest_subset:
+            table = manifest_subset[table_key_long]
+            name = table["name"]
+            schema = table["schema"]
+            database = table["database"]
+            source = table["unique_id"].split(".")[-2]
+            table_key = schema + "." + name
+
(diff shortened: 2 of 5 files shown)


bmtcril:
Exposures file looks like this:

Screenshot 2024-06-12 at 3 43 29 PM

bmtcril:
Exposure in the docs looks like this:

Screenshot 2024-06-12 at 3 46 38 PM

(If we fill in the dashboard descriptions in Superset they will propagate to the description there)

And lineage works: 
Screenshot 2024-06-12 at 3 48 09 PM

.github/workflows/integration-test.yml
```diff
@@ -76,6 +76,8 @@ jobs:
           tutor local do dump-data-to-clickhouse --options "--object course_overviews"
           make extract_translations
           tutor local do import-assets
+          tutor local do collect-dbt-lineage
+          tutor local run lms python manage.py lms transform_tracking_logs --source_provider LOCAL --source_config '{"key": "/openedx/data/", "prefix": "tracking.log", "container": "logs"}' --destination_provider LRS --transformer_type xapi
```
bmtcril, line 80: This is a somewhat random addition to this PR, but will test that we don't break the backfill command again

tutoraspects/templates/aspects/apps/superset/pythonpath/collect_dbt_lineage.py
```diff
@@ -0,0 +1,195 @@
+"""
```
bmtcril, line 1: A lot of this file is adapted from here: https://github.com/slidoapp/dbt-superset-lineage/blob/main/dbt_superset_lineage/pull_dashboards.py#L264

I thought I had documented that here, and will do so if this seems like a good direction.

pomegranited (approved):
:+1: Works great for me, thank you @bmtcril !

- [x] I tested this on my tutor dev aspects stack, and (after rebuilding the aspects-superset image) the command ran without issue.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] Includes documentation
- [x] <strike>User-facing strings are extracted for translation</strike> N/A

tutoraspects/templates/aspects/apps/superset/pythonpath/collect_dbt_lineage.py
```diff
@@ -0,0 +1,195 @@
...
+    aspects_dbt can add it to the project in CI and allow dbt docs to show
+    which models are being used in Superset."""
+)
+def collect_dbt_lineage():
+    """
+    Pull SQL from datasets, compare with dbt manifest.json to find used assets.
+    """
+    dbt_tables = get_tables_from_dbt()
+
+    target_dashboards = {{SUPERSET_EMBEDDABLE_DASHBOARDS}}
```
bmtcril, line 175: It occurs to me that I need to change this to include the Operator Dash as well.

Ian2012:
@bmtcril should we move this to `aspects-dbt` CI? Looks good to me, just it makes sense to have this exposed on the official DBT docs

bmtcril:
@Ian2012 The problem with putting this in aspects-dbt CI is that the exposures will be out of date and cause errors building the docs if tutor-contrib-aspects hasn't been updated to use them yet. If we do it in this repo we can get the correct versions of aspects-dbt & Superset dashboards.

Ian2012 (approved):

tutoraspects/templates/base-docker-compose-services
```diff
@@ -8,6 +8,8 @@ image: {{ DOCKER_IMAGE_SUPERSET }}
     - ../../env/plugins/aspects/apps/superset/scripts:/app/scripts
     - ../../env/plugins/aspects/build/aspects-superset/localization:/app/localization
     - ../../env/plugins/aspects/build/aspects-superset/openedx-assets:/app/openedx-assets
+    # Shared with Aspects job so it has access to the state file for the lineage push job
+    - ../../data/aspects-dbt:/app/dbt_manifest
```
bmtcril, line 12: @Ian2012 I'm pondering how to make this work with k8s based on our earlier discussion

bmtcril:
We're going to punt on k8s support for this temporarily and merge it as-is. Eventually we'd like to make the dbt docs using whatever the installed (potentially custom) version of dbt is and exposures from all (potentially custom) dashboards available in the Tutor environment.

PR #853, opened 2024-06-13
feat: Append uuid to filenames and rename existing files

saraburns1:
Append last 6 characters of uuid to all chart filenames to prevent file overwrites when there are duplicate chart names
Rename all existing charts & update dashboard chart IDs


tutoraspects/templates/aspects/build/aspects-superset/openedx-assets/assets/datasets/fact_instance_actors.yaml
```diff
@@ -52,5 +52,5 @@ schema: '{{ ASPECTS_XAPI_DATABASE }}'
 sql: ''
 table_name: fact_instance_actors
 template_params: null
-uuid: 5d7fa3b0-5692-5229-b8fc-4711b88d5401
+uuid: e78963de-0a12-5b15-a378-5bae3a8738e7
```
bmtcril, line 55: I'm not sure where these uuid changes are coming from, I can't find those new ones anywhere which is concerning.

saraburns1: all good 👍

tutoraspects/asset_command_helpers.py
```diff
@@ -225,15 +225,35 @@ class DatabaseAsset(Asset):
...
     """
     Check various aspects of the asset file based on its type.
 
     Returns the destination path for the file to import to.
+    Append last 6 characters of uuid for charts
     """
     orig_filename = os.path.basename(asset_path)
-    out_filename = re.sub(r"(_\d*)\.yaml", ".yaml", orig_filename)
-    content[FILE_NAME_ATTRIBUTE] = out_filename
+    out_filename_trimmed = re.sub(r"(_\d*)\.yaml", ".yaml", orig_filename)
```
bmtcril, line 246: Nit: can you move this into the `else` below so we're not having to do this unnecessarily?

tutoraspects/asset_command_helpers.py
```diff
@@ -243,12 +263,8 @@ def validate_asset_file(asset_path, content, echo):
...
             existing = None
 
-            # Check if the file already exists
-            if os.path.exists(os.path.join(out_path, out_filename)):
-                with open(
-                    os.path.join(out_path, out_filename), encoding="utf-8"
-                ) as stream:
-                    existing = yaml.safe_load(stream)
+            existing = _check_file_exists(out_path, out_filename_trimmed)
+            existing = _check_file_exists(out_path, out_filename_uuid)
```
bmtcril, line 267: I think we should only be doing one of these, and can maybe use `content[FILE_NAME_ATTRIBUTE]` for it?

saraburns1: I had it checking both for the initial renaming because it needed to find the original file to do the replacement but we probably don't need it anymore if we're going to rename everything all at once. I'll change it!

bmtcril (approved):
Worked for me when serializing the new course dashboard changes 👍

Ian2012 (approved):

PR #859, opened 2024-06-14
Bmtcril/limit ci languages

bmtcril:
Just curious how much CI might speed up with fewer languages / assets to test


Ian2012:
fyi, CI has been much slower since the v1.0.0 release as the docker cache has not been pushed. It should be push for PRs too, but not sure what's the problem

bmtcril:
Seems like it saves ~6 mins on the integration tests. 🤷

Ian2012 (approved):

PR #880, opened 2024-06-20
Force build version, add release instructions to docs

bmtcril:
This attempts to remove the ability to override the build version manually, which I don't think ever worked quite right. So we'll trust the automation going forward. Also adds a bit to the docs about how to release, since I always want to trigger the wrong workflow.

### DIFF
--- a/.github/workflows/bump-version.yaml
+++ b/.github/workflows/bump-version.yaml
@@ -1,13 +1,6 @@
 name: Bump version and changelog
 on:
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'The version to release (e.g 1.2.3)'
-        required: true
-        default: ''
-        type: string
-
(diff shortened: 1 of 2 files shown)


.github/workflows/bump-version.yaml
```diff
@@ -1,21 +1,14 @@
...
 on:
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'The version to release (e.g 1.2.3)'
-        required: true
-        default: ''
-        type: string
-
+  workflow_dispatch
```
bmtcril, line 3: This seems to be the way to do this without inputs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

Ian2012 (approved):

PR #881, opened 2024-06-21
fix: Import of custom Superset assets

bmtcril:
The custom Superset assets are all pushed into a single assets.yaml file, and need to be separated by `---` lines. This fixes the ability to read multiple custom assets from the assets.yaml / `superset-extra-assets` patch.

Also allows the `import_superset_zip` command to serialize assets to other locations so that extensions can use it as well.

### DIFF
--- a/tutoraspects/asset_command_helpers.py
+++ b/tutoraspects/asset_command_helpers.py
@@ -60,12 +60,12 @@ class Asset:
         if not self.path:
             raise NotImplementedError("Asset is an abstract class.")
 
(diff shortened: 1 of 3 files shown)


bmtcril:
@pomegranited you'll need this fix to use custom assets. My tests for this are here: https://github.com/bmtcril/tutor-contrib-aspects-extension

Ian2012 (approved):

bmtcril:
@Ian2012 this could use a re-review whenever you have a min.

Ian2012 (approved):

Issue #893, opened 2024-07-13
ch_report Authentication failed


CodeWithEmad:
I set up a new aspect plugin, on one of our orgs. but `ch_report` can't communicate with clickhouse.

![Screenshot from 2024-07-13 10-48-25](https://github.com/user-attachments/assets/852ad69b-fcc9-4767-9dc4-7e31f142cb41)

Aspect version: 1.0.2
Tutor version: 17.0.5

**Update**

After a bit of digging, I found out that the generated `ASPECTS_CLICKHOUSE_REPORT_PASSWORD` is not used in the init jobs, and changing the password to the value used in `.ci/config,yml` done the trick. :)
I'll open a PR if no one has done anything about this.

Ian2012:
This is already fixed but I'm not sure if on 1.0.3. Will review on monday

Ian2012:
Hi @CodeWithEmad. This was fixed and tested on dev and k8s environments: https://github.com/openedx/tutor-contrib-aspects/commit/4918d068506e16644790f9b85b6eee98c6a6b671

Can you leave more information of the environment you are running?

CodeWithEmad:
Thanks @Ian2012 for looking it up. https://github.com/openedx/tutor-contrib-aspects/commit/4918d068506e16644790f9b85b6eee98c6a6b671 looks more like fixing superset connection. If you're positive that ch_report's password issue has been fixed, I'll close this one. 

> Can you leave more information of the environment you are running?

Aspect version: 1.0.2
Tutor version: 17.0.5

Ian2012:
@bmtcril Have you experienced anything like this?

@CodeWithEmad You could try to rerun init jobs and check that the value of `ASPECTS_CLICKHOUSE_REPORT_PASSWORD` matches what is on the jobs

CodeWithEmad:
> You could try to rerun init jobs and check that the value of ASPECTS_CLICKHOUSE_REPORT_PASSWORD matches what is on the jobs

Well, I ran aspects for 3 organizations, and in all 3, the `ch_report` couldn't connect to click house. changing the password fixed the issue, 

> how are you installing the Aspects plugin?  Are you running Tutor dev, local, or k8s?

pip install 'tutor-contrib-aspects==1.0.2', in k8s mode.

bmtcril:
I'm working on this now, and am definitely seeing some oddities in a local k8s. The superset init job is throwing a lot of these, which we don't get on local:

```
[...]

PR #904, opened 2024-07-22
docs: Update readme to include the Superset image

bmtcril:
If this image is not built, the Superset database connections will have the wrong password and other settings changes will not be used.

Closes: #893


Ian2012 (approved):

CodeWithEmad (approved):
Thanks for keeping me in the loop.

PR #922, opened 2024-08-22
feat: Upgrade ClickHouse to 24.8

bmtcril:


saraburns1 (approved):

Ian2012 (approved):
I've changed the order in which the database tables were deleted to follow a child-parent order and removed an intermediate dictionary that was not deleted correctly.

PR #930, opened 2024-09-04
docs: correct the transform-tracking-logs do command name

Danyal-Faheem:
The `transform-tracking-logs` do command in the README was incorrectly spelled as `transform_tracking_logs`. I've updated this in the README.


Ian2012 (approved):

Ian2012:
@Danyal-Faheem thanks for the correction

PR #941, opened 2024-09-13
feat: Auto delete old charts

### LINKED ISSUE #370: Feat: Delete known-orphaned Superset assets

saraburns1:
Closes #370 

Output when importing zips
Screenshot 2024-09-13 at 10 47 38 AM
Screenshot 2024-09-13 at 10 47 53 AM

### DIFF
--- a/tutoraspects/asset_command_helpers.py
+++ b/tutoraspects/asset_command_helpers.py
@@ -502,6 +502,7 @@ def _get_all_dataset_uuids():
 
     return all_dataset_uuids
 
+
 def _get_all_chart_uuids():
     """
     Return the UUIDs of all datasets and charts in our file assets.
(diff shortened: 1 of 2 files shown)


bmtcril:
Nice, I'm glad you're tackling this! Once concern that I tried to raise in the original ticket is that we can't assume that any assets in Superset that aren't current Aspects assets should be deleted (like custom charts and dashboards) so we need some kind of way to say "this asset uuid is a deleted Aspects asset from the past that should be cleaned up".

saraburns1:
> Nice, I'm glad you're tackling this! Once concern that I tried to raise in the original ticket is that we can't assume that any assets in Superset that aren't current Aspects assets should be deleted (like custom charts and dashboards) so we need some kind of way to say "this asset uuid is a deleted Aspects asset from the past that should be cleaned up".

could we instead have a file that keeps a list of unused assets that we DONT want to delete? that might be easier to maintain than a list of ones TO delete

bmtcril:
I don't think we have a way of knowing which custom assets someone added are unused / deletable. We can only know our own assets, so we could put a check in CI or something that would fail a build if there were unused assets left in our changes. Then we could add them to the list. Seems kind of error prone, but if we put the reminder to add them to the list in the error message it might help?

saraburns1:
Ahhh makes sense. Cool, I'll do that

PR #945, opened 2024-09-16
DRAFT: Delete orphan assets

### LINKED ISSUE #370: Feat: Delete known-orphaned Superset assets

saraburns1:
Closes #370 

- UUIDs in aspects_asset_list.yaml that are not used in charts/dashboards will be deleted (orphaned) or ignored
- Fix 'used UUID' query to include datasets used in dashboards (for filters)
- Modify warnings/output to be more concise 
- *Split asset helper file to make more manageable


tutoraspects/asset_command_helpers.py
```diff
@@ -537,42 +174,102 @@ def _get_used_chart_dataset_uuids():
...
+                        delete_count += 1
+
+            for uuid, data in unused_chart_uuids.items():
+                if orphaned_uuids and orphaned_uuids.get("charts"):
+                    if uuid in orphaned_uuids.get("charts"):
+                        echo(f"Deleting orphan chart {data.get('name')} (UUID: {uuid})")
+                        os.remove(data.get("file_path"))
+                        delete_count += 1
+
+    echo(f"Deleted {delete_count} assets")
```
Ian2012, line 275: This would delete the files from the repository code but if an installation was running and old version with outdated assets those still would be created, you need to add a command for clean up as part of the initialization process

PR #950, opened 2024-09-25
feat: Remove unused aspects assets on import

### LINKED ISSUE #370: Feat: Delete known-orphaned Superset assets

saraburns1:
Closes #370

1. `create_assets.py` - run during init and import-assets
write to yaml file every uuid and it's translated asset uuids

2. `delete_assets.py` - called within create_assets
for each uuid in aspects_asset_list 'unused' and their corresponding translated uuids, DELETE assets from superset db

3. `delete_aspects_unused_assets.py` - run during 'import_superset_zip' and 'check_superset_assets'
if chart/dataset is not used AND is in aspects_asset_list under 'unused', DELETE yaml file
if chart/dataset is not used (ignoring any uuid listed under 'ignored' in aspects_asset_list.yaml) print warning about potential unused assets

4. Remove duplicate name check - not needed anymore since we append uuid to filename

image
image
image


tutoraspects/asset_command_helpers.py
```diff
@@ -474,19 +483,20 @@ def check_asset_names(echo):
...
-                        f"WARNING: Duplicate name {asset[k]} in {file_name}, this "
-                        f"could confuse users, consider changing it."
+                        f"WARNING: Duplicate {k} {asset[k]} in {asset.get('_file_name')}"
                     )
                 names.add(asset[k])
                 break
 
-    echo(f"{warn} duplicate names detected.")
+    echo(
+        f"{warn} duplicate names detected. This could confuse users, consider changing them."
```
bmtcril, line 498: I'm not sure how useful these warnings are these days, IIRC we had intentionally duplicated chart names between the individual learner and at-risk learner dashboards so maybe we should just get rid of this check?

saraburns1: That's fair, I'll remove it!

bmtcril (commented):
Code LGTM, can you update the PR description with exactly what the changes are and put something in the readme about the new command? I don't _think_ the old commands are in there anywhere but can you spin through and make sure?

tutoraspects/asset_command_helpers.py
```diff
@@ -518,10 +510,9 @@ def _get_used_chart_dataset_uuids():
             filters = asset["metadata"].get("native_filter_configuration", [])
 
             for filter_config in filters:
-                for filter_dataset in filter_config.get("target", {}).get(
-                    "datasetUuid", []
-                ):
-                    used_dataset_uuids.add(filter_dataset)
+                for item in filter_config.get("targets", {}):
+                    if item.get("datasetUuid"):
+                        used_dataset_uuids.add(item.get("datasetUuid"))
```
Ian2012, line 515: On superset 4.1.0 this will change to `dashboardId`, I didn't find how to match it against the `datasetUuid`. Anyway, problem for future me

tutoraspects/asset_command_helpers.py
saraburns1: oops I didn't see your comment @Ian2012 - it would still throw a warning for something like the course comparison dashboard where there are multiple charts named the same thing (but one is on course level and one is on course run level) There's nothing wrong/confusing about that, though, so I don't think we need the warning anymore.

README.rst
```diff
@@ -112,6 +112,11 @@ asset by editing the asset in Superset and selecting "Save As" to save it to a n
 # Note: If you are using custom assets you will need to rebuild your aspects-superset
 # image on your local machine with `tutor images build aspects-superset --no-cache`.
 
+Assets created for Aspects that are no longer used are specified in `aspects_asset_list.yaml`
+and will be deleted from the Superset database when import-assets is run.
```
bmtcril, line 116: We should add that `import-assets` gets run as part of init, since that's the way most people would interact with this.

tutoraspects/asset_command_helpers.py
saraburns1: What is changing to dashboardId? The dataset is definitely different from dashboard.

Ian2012: Sorry, `datasetId` I mean

tutoraspects/templates/aspects/apps/superset/pythonpath/aspects_asset_list.yaml
```diff
@@ -1,6 +1,10 @@
 unused_uuids:
   datasets:
   charts:
+    - b0e170ce-26f1-4452-a3c0-49b41fa6e529
+    - 8230a366-957d-4667-a2ff-3cbcb2f25f60
+    - 6b7610a7-41d0-4d14-be81-62a5a8f22d62
+    - b2fff8a7-3ace-4075-9a3f-5fa5b99f81bb
```
Ian2012, line 7: Can you comment the chart name only for reference?

tutoraspects/templates/aspects/apps/superset/pythonpath/delete_assets.py
```diff
@@ -0,0 +1,68 @@
...
+        logger.warning(uuid_list)
+
+        # Force our use to the admin user to prevent errors on delete
+        g.user = security_manager.find_user(username="{{SUPERSET_ADMIN_USERNAME}}")
+
+        try:
+            # Delete tagged object rows first
+            for id in id_list:
+                tag_command = DeleteTaggedObjectCommand(OBJECT_TYPES[asset_type], id)
+                tag_command.run()
```
Ian2012: This is not necessary as per Superset 4.0.2, and on Superset 4.1.0 this should happen automatically by the backend

tutoraspects/templates/aspects/build/aspects-superset/openedx-assets/assets/charts/Video_Engagement_8b0535.yaml
```diff
@@ -102,6 +102,6 @@ query_context:
   result_format: json
   result_type: full
 slice_name: Video Engagement
-uuid: 6b7610a7-41d0-4d14-be81-62a5a8f22d62
+uuid: 8b0535a8-a43f-49bf-9d50-439a16bd3f74
```
Ian2012: Usually we don't change the uuid of the assets.

README.rst
```diff
@@ -112,6 +112,11 @@ asset by editing the asset in Superset and selecting "Save As" to save it to a n
 # Note: If you are using custom assets you will need to rebuild your aspects-superset
 # image on your local machine with `tutor images build aspects-superset --no-cache`.
 
+Assets created for Aspects that are no longer used are specified in `aspects_asset_list.yaml`
+and will be deleted from the Superset database during `init` (specifically `import-assets`).
+Unused asset yaml files are deleted during `import_superset_zip` or by running the command
+`tutor aspects delete_unused_assets`.
```
bmtcril, line 118: I think this is currently incorrect? It doesn't seem to happen on init, and the command that seems to call it is `check_superset_assets`?

saraburns1: yeah i need to update it

tutoraspects/templates/aspects/apps/superset/pythonpath/delete_assets.py
saraburns1: It does happen automatically, but if there are multiple tag rows per object (which there are locally at least) then it fails. I created an issue in the superset repo

tutoraspects/templates/aspects/build/aspects-superset/openedx-assets/assets/charts/Video_Engagement_8b0535.yaml
saraburns1: sorry, this was a mistake from testing

tutoraspects/templates/aspects/apps/superset/pythonpath/create_assets.py
```diff
@@ -64,32 +70,47 @@ def create_assets():
...
-                process_asset(asset, roles)
+                process_asset(asset, roles, translated_asset_uuids)
+
+
+    for uuid in translated_asset_uuids:
+        translated_asset_uuids[uuid] = list(translated_asset_uuids[uuid])
+
+    # Write parent UUID & translated child UUIDs to yaml file
+    path = os.path.join(PYTHONPATH,"translated_asset_mapping.yaml")
+    with open(path, "w") as file:
```
bmtcril, line 90: Looks like this is making k8s sad since you can't write files there. Could you keep this is memory and just pass it to `delete_assets`, or am I missing something?

saraburns1: yup good call, ill do that

bmtcril (approved):
Worked for me! Thanks for all the work on this!

Ian2012 (approved):

PR #984, opened 2024-11-07
feat: Use new parameterized view 

saraburns1:
Query new parameterized view with org and course_name filters from superset


Ian2012 (changes requested):
Change the repository

Ian2012 (approved):

PR #995, opened 2024-11-27
fix: do not add mysql as a dependency when RUN_MYSQL=False

Danyal-Faheem:
Aspects would fail if someone was using an external MySQL server
We now only add MySQL as a dependency to containers if RUN_MYSQL=True


Ian2012 (approved):
Thanks for your contribution

PR #1018, opened 2025-01-30
feat: dbt model updates & cleanup for Aspects 2.0

saraburns1:
Use new dbt models from https://github.com/openedx/aspects-dbt/pull/141
GH issue [#127](https://github.com/openedx/aspects-dbt/issues/127)

Snapshot of all dashboards using old and new code (some numbers off by 1)
https://drive.google.com/drive/folders/1y5xIRhLxamUNUV5Qhet--HKbgmM1NhYV?usp=sharing

[Data comparison results](https://docs.google.com/spreadsheets/d/1ejyTLJ0d1KCnAoF260Kjwqmo-82zG0DpaRdvk6aSEfA/edit?gid=774748146#gid=774748146)
Sheet 1 - New model row counts match old (within <1%)
Sheet 2 - New models average memory usage decreased by about 9%, average superset time decreased by 3.5%

Testing instructions:
1. Config save
2. images build aspects aspects-superset
3. do init-clickhouse
4. tutor local do dbt -c 'run-operation remove_deprecated_models' --only_changed False
5. tutor local do dbt -c 'run --full-refresh --select +problem_events grading_events navigation_events enrollment_events' --only_changed False
6. tutor local do dbt --only_changed False
7. config save CLEAN
8. Import-assets

might need to delete ‘xapi.fact_instance_enrollments’ dataset in Superset if filters don't load


bmtcril:
I'm going to start testing this today by creating a large dataset on the current main branch and upgrading it then comparing the charts / data before and after. Aside from running the script to drop old models is there anything else special I should do to test?

saraburns1:
I don't believe so, but let me know if you run into any trouble

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0007_event_sink.py
```diff
@@ -79,6 +79,14 @@ def downgrade():
         "DROP TABLE IF EXISTS {{ ASPECTS_EVENT_SINK_DATABASE }}.course_blocks"
         f"{on_cluster}"
     )
+    op.execute(
```
bmtcril, line 82: I think this needs to be in a new migration otherwise existing installs won't get it since they've already run this migration.

tutoraspects/templates/aspects/apps/aspects/migrations/alembic/versions/0041_drop_course_dictionaries.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    op.execute(
```
bmtcril, line 23: These are only being dropped on downgrade? I thought in the previous version we wanted them gone in the current state?

saraburns1: Nope, these are new views that reference `course_overviews` so when that table is dropped in the downgrade of 0007, these need to be dropped first

tutoraspects/plugin.py
```diff
@@ -384,8 +384,8 @@
         # DBT_PROFILE_* settings get passed into the dbt_profile.yml file.
         # For now we are pulling this from github, which should allow maximum
         # flexibility for forking, running branches, specific versions, etc.
-        ("DBT_REPOSITORY", "https://github.com/openedx/aspects-dbt"),
-        ("DBT_BRANCH", "v3.33.11"),
+        ("DBT_REPOSITORY", "https://github.com/saraburns1/aspects-dbt"),
```
bmtcril, line 387: Just need to update these when aspects-dbt releases.

bmtcril (approved):
Just the one remaining task. When it comes time to release let's do it together since the process is so finicky.

bmtcril (changes requested):
Actually, can you change this commit message from "feat:" to "feat!:" to make sure the versioning automation picks this up as a breaking change and updates the version appropriately?

bmtcril (approved):

PR #1019, opened 2025-01-31
feat: add course name to run charts

saraburns1:
Closes https://github.com/openedx/openedx-aspects/issues/301
Also added Org to the same charts on the 'Course Metrics' tab

image
image
image
image


bmtcril:
Have you checked how these look with a real (long) course name / org? Just want to make sure it doesn't break UI

saraburns1:
Squishes the graph a bit but I think it still looks fine
image

bmtcril (approved):
LGTM

PR #1048, opened 2025-04-04
feat: Upgrade ClickHouse to 25.3

bmtcril:
### This PR is in testing!

This is the latest LTS from CH, and includes a number of performance improvements, bug fixes, and features including refreshable materialized views.


bmtcril:
I believe we found that with additional testing the differences were within the normal bounds of operation. Is that right @saraburns1 ?

saraburns1 (approved):

PR #1051, opened 2025-04-08
chore: django 5.2 upgrade

saraburns1:
Ran test_tracking, load-xapi, import to Superset. 
Took action in studio & lms, events successfully entered in Clickhouse
All tests pass


bmtcril (approved):
🚀

PR #1067, opened 2025-05-06
fix: use end of block id for in-context filtering

saraburns1:
The In-Context front end only passes the last hash string of the block id to the dashboards for filtering

block-v1:course-v1:Org0+DemoX+2bc51b+type@problem+block@**42e10810**


bmtcril (approved):
Seems ok, none of those queries are used by non-in-context charts?

saraburns1:
> Seems ok, none of those queries are used by non-in-context charts?

The only query that is used by a non-in-context is watched_video_segments but none of the charts that use it use the block_id field

PR #1070, opened 2025-05-12
fix: Incontext labels

saraburns1:
Use new dbt models
Change xaxis labels to name and location
Fix links in embedded dashboards

closes https://github.com/openedx/openedx-aspects/issues/321
closes https://github.com/openedx/openedx-aspects/issues/318


tutoraspects/plugin.py
```diff
@@ -342,25 +347,7 @@
         (
             "SUPERSET_DASHBOARD_LOCALES",
             [
-                "ar",
```
bmtcril, line 345: These changes will need to be reverted, right?

saraburns1: yup

bmtcril (commented):
Looks good once aspects-dbt is released and updated

PR #1075, opened 2025-05-28
WIP Video updates

saraburns1:


bmtcril (commented):
LGTM once the dbt version is bumped. That should be a major version bump since it's changing the schema significantly.

Issue #1088, opened 2025-07-14
Mismatched required Tutor versions in base.in


bmtcril:
Currently we are pinning Tutor to >=18, since everything in Aspects should function back to Redwood with the exception of in-context analytics. However as part of the in-context work we pinned tutor-mfe to >=19 which effectively pins us to Sumac instead.

I think the answer is to change the tutor-mfe pin back to >= 18 and document that in context doesn't work before then, but I'm not sure if that would cause errors or require us to code some guardrails around it. I've reached out to Farhaan about it.

farhaanbukhsh:
Hey @bmtcril  thank you for raising the issue and working on the PR, there is one issue though if you see the comparison between [tutor-mfe 18.1.0 and 19.0.0](https://github.com/overhangio/tutor-mfe/compare/v18.1.0...v19.0.0) we can see that course-authoring is renamed to authoring specifically  and this will cause issues, hence depending of version of tutor-mfe we need to change course-authoring to authoring and vice versa. Yeah so in-short for in-context metrics to work we need tutor-mfe v19 and above. Hope that clarifies things and your PR should technically working depending on the user version of dependencies.

PR #1089, opened 2025-07-14
Lower Tutor MFE version pin

bmtcril:
Allow this plugin to be installed on Redwood and Sumac again. Previously the MFE version pin forced a Tutor upgrade to Teak or higher. Documentation updated to include more references to version compatibility.

Closes: #1088


tutoraspects/plugin.py
```diff
@@ -683,111 +687,114 @@ def _mount_superset_compose(
...
-                priority: 60,
-                type: DIRECT_PLUGIN,
-                RenderWidget: SubSectionAnalyticsButton,
-            },
-          }""",
-        ),
-    ]
-)
+# If PLUGIN_SLOTS doesn't exist, we are on Redwood and do not
+# support in-context metrics.
```
farhaanbukhsh, line 691: and tutor-mfe version needed to support in-context metric is v19 and above.

saraburns1 (approved):

tutoraspects/plugin.py
farhaanbukhsh: @bmtcril  Let me try to understand what we are trying to do. We will create a toggle in platform-plugin-aspects, which opens up a toggle in-context API that determines if in-context metrics will be displayed or not.

Why do we need this?

The feature is still toggleable by ASPECTS_ENABLE_STUDIO_IN_CONTEXT_METRICS, but if you want to do it at the env.config.js level or  PLUGIN_SLOTS level, maybe we can add a gate here, or we can eliminate the use of PLUGIN_SLOTS.add_item by writing directly to https://github.com/openedx/tutor-contrib-aspects/blob/643aeec1e56c3fc3e0851f34740b241de07ad448/tutoraspects/patches/mfe-env-config-runtime-definitions-authoring

But these will be done at build time and not at runtime.

Let me know if I am not able to understand something.

farhaanbukhsh: Now that I think more about it, making this a runtime toggle solution doesn't seem to be straightforward.  The decision needs to be taken in env,config.js or inside the plugin slots which doesn't seem to be a good solution.

farhaanbukhsh: @bmtcril  sounds right :) thank you for the amazing work.

Issue #1091, opened 2025-07-22
Plugin causes React Query peer conflict (v4 vs v5) — breaks further npm install


ihor-romaniuk:
Installing `@openedx/frontend-plugin-aspects` into the `frontend-app-authoring` micro-frontend with `--legacy-peer-deps` succeeds, but **locks the dependency graph**. Any subsequent `npm install` fails with a peer dependency conflict between `@tanstack/react-query@4` and `@tanstack/react-query@5`.

```
{% if ASPECTS_ENABLE_STUDIO_IN_CONTEXT_METRICS %}
   RUN --mount=type=cache,target=/root/.npm,sharing=shared npm install --legacy-peer-deps openedx/frontend-plugin-aspects
{% endif %}
```

### Steps to Reproduce

1. Install `tutor-contrib-aspects` plugin with enabled `ASPECTS_ENABLE_STUDIO_IN_CONTEXT_METRICS` flag
2. Add any additional NPM package to frontend-app-authoring via post npm install process 
3. Observe failure:
```
npm ERR! While resolving: @openedx/frontend-plugin-aspects@0.1.0
npm ERR! Found: @tanstack/react-query@4.36.1
npm ERR! Could not resolve dependency:
npm ERR! peer @tanstack/react-query@"5.83.0" from @openedx/frontend-plugin-aspects@0.1.0
```

### Expected Behavior
- Plugin should not block further package installations.
- Peer dependencies should remain compatible or clearly documented.

### Actual Behavior
- Any new `npm install` command fails due to unresolved peer dependency conflicts.
- NPM tries to re-evaluate the full tree and breaks due to incompatible react-query versions.

### Temporary workaround on our side
We patched the Docker install order to assign a higher priority to the frontend-plugin-aspects install, ensuring it runs last:
```
if patch_file.name.endswith('mfe-dockerfile-post-npm-install-authoring'):
    hooks.Filters.ENV_PATCHES.add_item((os.path.basename(path), patch_file.read()), 100)
else:
    hooks.Filters.ENV_PATCHES.add_item((os.path.basename(path), patch_file.read()))
```

bmtcril:
@farhaanbukhsh @tecoholic would you mind taking a look?

tecoholic:
@farhaanbukhsh From the [issue implicitly linked by Github](https://github.com/raccoongang/tutor-contrib-aspects/pull/1), I think the `--legacy-peer-deps` used here affects the MFE when other plugins are added.

[...]

PR #1093, opened 2025-07-30
fix: Removes legacy peer dependency flag to make the install flexible

farhaanbukhsh:
The changes remove legacy-peer-dependency flag because we matched the dependency in frontend-plugin-aspects, this helps to be flexible with installing dependencies as reported in #1091.

**Discussions**: 
https://github.com/openedx/frontend-plugin-aspects/pull/51

**Dependencies**: None

**Screenshots**: Always include screenshots if there is any change to the UI.

**Testing instructions**:

Same as https://github.com/openedx/frontend-plugin-aspects/pull/51


bmtcril:
@farhaanbukhsh can you throw a bit more context in the description? Is this related to / will it close https://github.com/openedx/tutor-contrib-aspects/issues/1091 ?

farhaanbukhsh:
> @farhaanbukhsh can you throw a bit more context in the description? Is this related to / will it close #1091 ?

Sorry for the blank description I have added more context now.

bmtcril (approved):

PR #1094, opened 2025-07-31
feat: Trying out Altinity lightweight connector

bmtcril:
This is a **highly** speculative attempt to replace the platform-plugin-aspects sinks with something more robust and reliable. Currently this test works, creating ClickHouse tables to match MySQL versions and appending data as MySQL changes (for tables set in `ALTINITY_SINK_TABLE_LIST`). It is not plumbed through dbt or into downstream reports at this time.

Currently only Tutor local should work, dev is unlikely to work, k8s is not implemented.


bmtcril:
We're not pursuing this right now

PR #1100, opened 2025-08-13
Engagement updates

saraburns1:
Use new dbt models from https://github.com/openedx/aspects-dbt/pull/156
Cleanup charts and datasets

Ran performance-metrics before and after these changes on aspects-beta:
Avg Memory decreased by 25% (16.46 MB vs 12.31 MB)
Avg Superset Load Time decreased by 12% (0.45s vs 0.40s)


bmtcril (approved):
Looks good! Testing with a 100M xapi dataset showed overall 12% speedup 5% memory reduction using the performance metrics script.

Issue #1103, opened 2025-08-18
Update Superset worker healthcheck


saraburns1:
From https://discuss.openedx.org/t/tutor-contrib-aspects-plugin-superset-worker-healthcheck-fail/16760

Celery 5.0 changed the syntax for the `inspect` command so we need to update the docker patch files.

[tutoraspects/patches/local-docker-compose-services](https://github.com/openedx/tutor-contrib-aspects/blob/5f7d747f4b0376b567ee3ff013f56bc775897f51/tutoraspects/patches/local-docker-compose-services#L68)
[tutoraspects/patches/local-docker-compose-dev-services](https://github.com/openedx/tutor-contrib-aspects/blob/5f7d747f4b0376b567ee3ff013f56bc775897f51/tutoraspects/patches/local-docker-compose-dev-services#L23)

change `celery inspect ping -A superset.tasks.celery_app:app -d celery@$$HOSTNAME"`
to `celery -A superset.tasks.celery_app:app inspect ping -d celery@$$HOSTNAME"`

PR #1104, opened 2025-08-20
fix: Superset Celery issues with healthcheck and pid

bmtcril:
Superset worker's healthcheck was incorrectly formatted for the version of Celery that Superset now uses.

Superset beat was failing to start at all because the image already had a pidfile on it. We now delete that if it exists.

Tested by taking down redis until the Supserset worker showed unhealthy.

Closes: #1103


tutoraspects/templates/aspects/apps/superset/docker/docker-bootstrap.sh
```diff
@@ -43,6 +43,8 @@ if [[ "${1}" == "worker" ]]; then
   echo "Starting Celery worker..."
   celery --app=superset.tasks.celery_app:app worker -Ofair -l INFO
 elif [[ "${1}" == "beat" ]]; then
+  echo "Removing any orphan pid..."
+  rm -rf /tmp/celerybeat.pid
```
saraburns1, line 47: why is this being removed?

PR #1150, opened 2025-12-09
fix: Add user to Clickhouse compose file

bmtcril:
This seems to be necessary for rootless Docker and possibly Podman setups.


bmtcril:
These settings seem mutually exclusive with the default Docker setup, so I'm closing this for now

PR #1182, opened 2026-02-09
chore: Update tests to use org and course name

bmtcril:
We found a bug that was missed by our smoke tests due to not having the superset filters. Trying to add them here.


saraburns1:
Included in #1186 - closing this

PR #1183, opened 2026-02-10
fix: Remove 'Course Name' field from at_risk dataset

saraburns1:
Removed 'Course Name' field from dataset configuration.


bmtcril (commented):
I think we may need this change in problem_coursewide_avg.yaml as well to deal with the individual learner dash? I'm not sure if that will create problems in the course dash, though...

PR #1224, opened 2026-04-13
chore: Replace kube-tools with arkade-get in workflow

saraburns1:
kube-tools is deprecated and was failing with recent req updates


.github/workflows/integration-test.yml
```diff
@@ -187,7 +187,7 @@ jobs:
           pip install -e .
           tutor config save
       - name: Run Kubernetes tools
-        uses: stefanprodan/kube-tools@v1
+        uses: alexellis/arkade-get@master
```
bmtcril: Can we ping this to a commit or tag?

PR #1233, opened 2026-04-23
feat: Upgrade to Superset v6

saraburns1:
Upgrades superset
Includes changes from @Ian2012 to performance metrics and aspects-dbt mount

### DIFF
--- a/tutoraspects/templates/aspects/build/aspects-superset/Dockerfile
+++ b/tutoraspects/templates/aspects/build/aspects-superset/Dockerfile
@@ -4,7 +4,7 @@
 # https://github.com/apache/superset/blob/master/Dockerfile
 # https://superset.apache.org/docs/databases/installing-database-drivers
 
-FROM apache/superset:4.1.1
+FROM apache/superset:6.0.0
 
 USER root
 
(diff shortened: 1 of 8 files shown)


tutoraspects/templates/aspects/build/aspects-superset/Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN apt-get update -qq \
     && ln -s /opt/firefox/firefox /usr/local/bin/firefox \
     && apt-get autoremove -yqq --purge wget bzip2 && rm -rf /var/[log,tmp]/* /tmp/* /var/lib/apt/lists/*
 
+ENV PATH="/app/.venv/bin:${PATH}"
 RUN pip install -r /app/requirements.txt
```
bmtcril: Based what I'm seeing [here](https://github.com/apache/superset/blob/6.0.0/Dockerfile#L172) I think this should be:

```
RUN --mount=type=cache,target=${SUPERSET_HOME}/.cache/uv \
    uv pip install -r /app/requirements.txt
```

bmtcril (commented):
Looks good. Last think I can think of, have you tested all of the commands? I think we test most of them in CI, but I'm not sure we get them all

saraburns1:
i tested:
import-assets
check_superset_assets
import_superset_zip
performance-metrics
init -l aspects

any others you can think of?

bmtcril:
> i tested: import-assets check_superset_assets import_superset_zip performance-metrics init -l aspects
> 
> any others you can think of?

They should all be in commands_v1.py:

- `load_xapi_test_data` (should be tested in CI)
- `dbt`
- `alembic`
- `init_clickhouse`
- `collect_dbt_lineage`
- `dump_data_to_clickhouse`
- `transform_tracking_logs` (maybe tested in CI?)

saraburns1:
[Untitled.rtf](https://github.com/user-attachments/files/27056345/Untitled.rtf)
cool - all successful!

bmtcril (approved):

PR #1238, opened 2026-04-27
feat! Update Python version requirement to 3.12

saraburns1:
Need to change this to force version 3.12 or else platform-plugin-aspects newest version won't work


bmtcril:
Weird, I'm trying to think of where we import platform-plugin-aspect in Tutor. Do you have the errors you were seeing on hand? I would only expect it to need to be installed in openedx-platform, but maybe I'm forgetting something.

PR #1245, opened 2026-05-06
fix: fixes for superset 6

saraburns1:
update dbt version
add new role permissions for themes in embedded dashboards
fix issue with adding extra tags when uploading dashboards


bmtcril (approved):
👍

PR #1207, opened 2026-03-20
feat: set vector as default data pipeline

Ian2012:
This PR sets Vector as the default data pipeline. It also includes a couple of improvements:

- Stores the alembic migration in the same database to avoid conflicts (thanks @bmtcril).
- Add tutor mounts for aspects-dbt. Now Aspects developers can work with a local copy of aspects-dbt without needing to push their changes to github. Just run `tutor mounts add ./aspects-dbt/` or the directory where aspects-dbt is stored and run your local copy.
- Enables Vector by default.

> [!CAUTION]
> This is a breaking change. Users which install this version will disable their Ralph workloads if they do not update their configuration.

Depends on: https://github.com/openedx/aspects-dbt/pull/164

Fixes: https://github.com/openedx/tutor-contrib-aspects/issues/1126 https://github.com/openedx/tutor-contrib-aspects/issues/1096

Issue #1096, opened 2025-08-04
Change default event pipeline


bmtcril:
After some discussion with operators running Aspects in production, we have heard that the current default pipeline with the Ralph LRS handling ClickHouse inserts and providing the xAPI interface is proving to be a performance bottleneck and the xAPI features are under utilized.  Operators have been able to achieve high Ralph throughput by:

- Running more Ralph containers 
- Running more Celery workers
- Changing some hard coded values and behavior in a Ralph fork

Vector seems to handle these things loads automatically without the need to scale to more containers and do ClickHouse batching based on volume without operator intervention.

Therefore in our next set of breaking changes we would like to change from the default data pipeline from Ralph to Vector. There would, however, still potentially need to be more Celery workers since that is where the events are transformed.

This should involve the following:

- Set [RUN_VECTOR](https://github.com/openedx/tutor-contrib-aspects/blob/b610d1f738d775209093a54d698e11ec996f2178/tutoraspects/plugin.py#L36) to True
- Set [RUN_RALPH](https://github.com/openedx/tutor-contrib-aspects/blob/b610d1f738d775209093a54d698e11ec996f2178/tutoraspects/plugin.py#L38) to False
- Updating documentation and release notes to indicate the change (and steps needed to maintain the current pipeline on an upgrade)
- Perform before and after performance tests to confirm our assertions that Vector can handle the same or greater volume of events that Ralph can, without changes on the operator side
- Write up a performance evaluation about how Vector scales and how it performs compared to Ralph
- Create issues for any problems that come from testing the Vector implementation

PR #1250, opened 2026-05-19
Set Vector as default backend

### LINKED ISSUE #1126: Bug: Migrations for Vector break when using it as the xAPI source
For testing purposes we've maintained a separate set of migrations for the `xapi_events_all` table in Vector's `openedx` ClickHouse database which allows us to run both the Celery / event bus pipeline and Vector pipeline at the same time and compare the data, and allows Vector permissions to be limited to just that database. However, in order to use the Vector table as the source of truth for the rest of Aspects we need to change the `ASPECTS_XAPI_DATABASE` setting from it's default of `xapi` to `openedx`.

When we do this, the `xapi_events_all` migrations all try to operate on the same table in `openedx` causing collisions and errors (specifically in migration 0011). This has rendered the Vector configuration unusable for some time.

When migrations run to completion using the default settings, the `xapi_events_all` tables in `openedx` and `xapi` are structurally identical.

I can see 3 potential ways forward:

## 1. Consolidate xAPI in one table
- Drop the `openedx` version of the tables and have all paths write to a single table in `xapi`.
- This simplifies migrations and configuration at the cost of not being able to run both pipelines at the same time. It also means that we're always writing xAPI to the database with the name that makes sense for that, and we won't have extra, probably empty, versions of that table lying around.
- We would also need to grant write access to Vector there.

## 2. Separate config variable for dbt
- Right now dbt's use of the `ASPECTS_XAPI_DATABASE` setting to determine where to look for events is what's preventing us from simply creating the table in the right place. 
- Creating a new variable that dbt can use to choose between the source tables [here](https://github.com/openedx/aspects-dbt/blob/main/models/base/sources.yml) would allow us to continue to create both tables and therefore load the same data into both and test, while untangling the migrations... at the cost of even more configuration.
- There is also an issue with dbt always using the [ASPECTS_VECTOR_RAW_XAPI_TABLE](https://github.com/openedx/aspects-dbt/blob/main/models/base/sources.yml#L9C34-L9C63) setting to find the table name, which may not always be correct, that we could try to fix as part of this as well.

## 3. Branch in migrations
- We could skip the Vector migrations if `ASPECTS_XAPI_DATABASE == ASPECTS_VECTOR_DATABASE`, which would have the effect of creating `xapi_events_all` only in the `openedx` database.
- This would allow migrations to run at the cost of not being able to run both pipelines simultaneously.
- Additionally, switching configurations to even try to use the Celery/event bus pipelines would result in a broken schema with no recourse.

I'm leaning towards 2 since it has the fewest moving pieces and simplifies the testing story without having to muck with migrations, but I think 1 is viable and more intuitive. @saraburns1 what do you think?

bmtcril:
This is a rebase of https://github.com/openedx/tutor-contrib-aspects/pull/1207 updated for post-Verawood. It replaces the Celery data pipeline with Vector as default, which should provide a lot of performance and operational benefits.

- Stores the alembic migration in the same database to avoid conflicts
- Add tutor mounts for aspects-dbt. Now Aspects developers can work with a local copy of aspects-dbt without needing to push their changes to github. Just run tutor mounts add ./aspects-dbt/ or the directory where aspects-dbt is stored and run your local copy.
- Enables Vector by default

> [!CAUTION]
> This is a breaking change. Users which install this version will disable their Ralph workloads if they do not update their configuration.

Depends on: https://github.com/openedx/aspects-dbt/pull/164

Fixes: https://github.com/openedx/tutor-contrib-aspects/issues/1126 https://github.com/openedx/tutor-contrib-aspects/issues/1096


bmtcril:
@farhaanbukhsh this should be ready for testing when you get a chance!

saraburns1 (approved):
LGTM - tested locally

events in vector: 
image

superset data using openedx schema:
image

farhaanbukhsh:
Thank you @saraburns1  for reviewing it and sorry @bmtcril I couldn't get to it sooner.

PR #1305, opened 2026-08-21
feat: upgrading to superset 6.1

saraburns1:
in context metrics work:
image

reports tab works:
image

-------------

The `CurrentUserRestApi` permission is needed now due to stricter security introduced in https://github.com/apache/superset/pull/36410 (v6.0.1) and then a followup fix to force Public roles to define the necessary permission (https://github.com/apache/superset/pull/38474)


bmtcril (approved):
Nice! 🎉

PR #1309, opened 2026-08-25
fix: custom dark theme for superset 6.1

### LINKED ISSUE #1281: bug: White text on light backgrounds in Superset dark mode
Superset 6 dark mode changes the text color to be white, so some charts on our default color scheme end up with white text on light backgrounds. We should use the fun new design system to override that on dark mode or make the colors darker:

Image

saraburns1:
- Custom color schemes for default and dark themes - Superset v6.1 adjusts text color based on background (changes to black if color is too light) which fixes #1281 

image
image

- updating all dashboards to use Aspects color scheme
(before)
image
(after)
image

- Lots of superset yaml additions from upgrade to 6.1


bmtcril (approved):
Just curious, was `Initial_Numeric_Responses` unused?

saraburns1:
@bmtcril  - yes, we were planning to use it in In-Context metrics but didn't because we were hoping to show either the text response table OR this table depending on the type of problem and that isn't possible in Superset right now

PR #1310, opened 2026-08-26
feat: add vector aggregator

bmtcril:
Rebased version of #1215 

This PR changes the architecture to use Vector Agent + Vector Aggregator, reducing the risk of Too many parts errors from ClickHouse.

Small and medium sized k8s installation should see a reduced amount of writes to ClickHouse when switching either from Ralph or from the old Vector agent setup. Events are realtime collected from k8s nodes and sink to the k8s aggregatorm which then parses, processes and send them to ClickHouse and S3 (if enabled).

Large sized installations should see a reduction in ClickHouse disk usage and CPU usage as writes will be batched more efficiently from the aggregators instead that each agent concurrently writing to ClickHouse


saraburns1:
LGTM - ran 'load-xapi-test-data' locally with k8s and confirmed all data went through vector and into clickhouse

saraburns1 (approved):

PR #1317, opened 2026-09-10
feat: Hide specified charts when PII is disabled

saraburns1:
New functionality to hide specific charts when PII is disabled. 

On import, chart type is set to '' which removes the chart from the superset dashboard while keeping the original code. If PII is enabled or the list of uuids to hide is empty, the chart type is set back to CHART which shows the chart in the dashboard again. 

ASPECTS_ENABLE_PII: true
image
ASPECTS_ENABLE_PII: false
image


bmtcril (approved):
LGTM!

